PR #71, opened 2015-02-07
Rework of mutation and growth requirement.

Imasius:
Pull request for discussing the mutation / growth requirement rework. Please do **not merge** until it is
finished.

What was done so far:
- Moves growth / mutation requirements into a new class
- Removes required block from Mutation class.
- Soil whitelist is initialised with Blocks.farmland
- Refactor of "canGrow", "canMutate", "validSoil", .... methods into one place (they exist multiple times)

What needs to be done:

- Enhance MineTweaker3 support so mutations can easily be managed
- ~~(optionally) remove the mutation txt config~~
- ~~(optionally) if mutation txt config is removed, we need to find a way for users to list all mutations in the system~~
- (optionally) add NEI GUI for possible products of a plant


src/main/java/com/InfinityRaider/AgriCraft/farming/GrowthRequirement.java
```diff
@@ -0,0 +1,193 @@
...
+
+    public static final GrowthRequirement DEFAULT = new GrowthRequirement();
+    public static HashMap<ItemSeeds, HashMap<Integer, GrowthRequirement>> overrides = new HashMap<ItemSeeds, HashMap<Integer, GrowthRequirement>>();
+
+    /** Maximum allowed brightness, exclusive **/
+    private int maxBrightness = 16;
+    /** Minimum allowed brightness, inclusive **/
+    private int minBrightness = 8;
+
+    private static List<BlockWithMeta> defaultSoils = new ArrayList<BlockWithMeta>();
```
Imasius: Whats the difference between this list and the SoilWhitelist?

src/main/java/com/InfinityRaider/AgriCraft/farming/GrowthRequirement.java
```diff
@@ -0,0 +1,193 @@
...
+    private static List<BlockWithMeta> soils = new ArrayList<BlockWithMeta>();
```
Imasius: Why is the list static now? By specifying only a single block we have again different cases when we have multiple possible soils.

Imasius:
I splitted the GrowthRequirement into 2 classes. Feel free to rename the other class. Reason is, that GrowthRequirement currently has 2 responsibilities. Representing a requirement for a specific plant + utility methods, and managing all requirements and all soils.

PR #79, opened 2015-02-14
Add villagers

Elec332:


InfinityRaider:
What textures do these villagers use?

Elec332:
https://github.com/InfinityRaider/AgriCraft/blob/master/src/main/java/com/InfinityRaider/AgriCraft/proxy/ClientProxy.java#L97

PR #89, opened 2015-02-17
Update fr_FR.lang

Mazdallier:


InfinityRaider:
As always, thanks

Mazdallier:
This is not a big deal. ^^

Issue #130, opened 2015-03-12
AppleCoreHelper May Not See AppleCore as Loaded (Possibly Intermittent Issue)


crawford42:
While testing a private mod pack last night, I noticed that potato and carrot seeds grew at the Vanilla rate rather than the Hunger Overhaul speed. Everything I could find in forums, etc. said it should Just Work, so I started digging into the source.

I believe the static initialization in AppleCoreHelper is being called before the FML Loader has loaded/registered AppleCore, and is getting an incorrect answer as to whether AppleCore is loaded. Static blocks and variable initializations are called when the class is loaded; there's no guarantee as to when that might happen. Sometimes it might be OK, sometimes it won't. It may depend on the order in which the JAR files were added to the mod directory!

I recommend moving this initialization to a static method and calling that method from AgriCraft.postInit, so it's done at a stage when the FML should have AppleCore loaded.

I'll try to have a pull request for this tonight or tomorrow.

crawford42:
Looking a bit more, I see the ModIntegration class is a better place to call it from.

InfinityRaider:
You will PR this? okay thanks, I won't touch it then, also please PR it to 1.3-bugfix, thanks.

PR #131, opened 2015-03-13
#130 -- Change timing of check for AppleCore API

crawford42:
The ModIntegration class now calls AppleCoreHelper.init, rather than AppleCoreHelper depending on static initialization.


crawford42:
Sorry -- messed up the pull request.

PR #132, opened 2015-03-13
#130 -- Change timing of check for AppleCore API

crawford42:
ModIntegration calls AppleCoreHelper.init rather than AppleCoreHelper depending on static initialization.


InfinityRaider:
Cheers (Closes #130)

PR #154, opened 2015-04-03
API

HenryLoenwind:
My take on an API.

Please note that the implementation is untested. I did not want to write code that uses the API before its interface was finalized.


InfinityRaider:
Can you PR it into the API branch please?

HenryLoenwind:
sure

PR #155, opened 2015-04-04
API

HenryLoenwind:
My take on an API.

Please note that the implementation is untested. I did not want to write code that uses the API before its interface was finalized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,10 @@ schematics/
 AgriCraft_Dependencies.rar
 .idea/
 logs/
+/bin/
+
+.classpath
+.project
+.settings/
+dependencies.zip
+*.bak
--- a/build.gradle
+++ b/build.gradle
@@ -148,3 +148,12 @@ sourceSets {
         }
     }
 }
+
+task apiJar(type: Jar) {
+    from sourceSets.main.allSource
+    from sourceSets.main.output
+    include 'com/InfinityRaider/AgriCraft/api/**/*'
+    classifier = 'api'
+}
+
+build.dependsOn apiJar
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -1,4 +1,18 @@
 #!/usr/bin/env sh
 
-wget -O ./dependencies.zip "https://www.dropbox.com/s/kasdmtxazgg5pth/dependencies.zip?dl=1"
-jar xvf ./dependencies.zip
+wget --no-check-certificate -O ./dependencies.zip "https://www.dropbox.com/s/kasdmtxazgg5pth/dependencies.zip?dl=1"
+
+cygwin=false
+case "`uname`" in
+  CYGWIN* )
+    cygwin=true
+    ;;
+esac
+
+zippath=./dependencies.zip
+# For Cygwin, ensure paths are in UNIX format before anything is touched.
+if $cygwin ; then
+    zippath=dependencies.zip
+fi
+
+jar xvf $zippath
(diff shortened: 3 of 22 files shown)


InfinityRaider:
Thanks, this looks really well done. Also it would have probably taken ages for me how to figure out to actually do an api and this gives me something to learn from and build forth onto.

HenryLoenwind:
You're welcome. And I actually need that API for Ender IO ;)

Is it clear how to extend that API after v1 has been published? (Add new interface to the api package, add new implementation to the apiimpl package (or extend current implementation to conform to both versions), return api object according to requested api version) And please, never(!) *change* any classes/interfaces in the api package after they have been published. Add new api versions, and drop support for old ones as you like, but any change will lead to crashes.

BTW: The api version number in the package-info.java does not need to be changed as long as you use the getAPI(version) construct. In fact, it should not be changed, as it stops older mods working with never AgriCraft version even if the old API they use is still supported.

InfinityRaider:
I'll just do a new version with the SeedRequirements class replaced by the GrowthRequirement class. I've also merged your api into my cropRewrite branch to fix any incompatibilities the rewrite might have introduced. One more thing I'll do is move some classes over to the api.

InfinityRaider:
Okay i've merged your api code into my CropRewrite branch (https://github.com/InfinityRaider/AgriCraft/compare/CropRewrite). This will be the code for the next version, so if you build onto anything please do it witht he code on that branch.

Also one question, can there be compiled code in the api package or should there only be interfaces there?

HenryLoenwind:
I think the code is needed. The api jar will be present while compiling a mod, and Java likes to compile again code, not source.

InfinityRaider:
Well okay, if I ever do something that I shouldn't do I'm sure I'll hear it :p.
And again, I really appreciate your effort.

HenryLoenwind:
So I finally had some time to start using the API.

(1) The dependencies.zip in the cropRewrite branch is incomplete Took me a while to get that sorted out. Having two versions of a mod in the libs folder didn't help at all ;).

(2) In my opinion, having classes in the api package is a bad idea. Reasons:

* If people ship the API with your mod by accident it is no problem with unchanging interfaces (as I set it up). If there are any classes in there, it will break as soon as there are different versions of one class in existence.

* Exposing you code that way can lead to people coding after the code, not the API documentation. You'd be forced to keep your old bugs around because others rely on them. Sort of what Microsoft has been doing with Windows for the last 30 years.

* Any class in the API that references a class outside the API breaks the API. At the moment that's CropPlant and GrothRequirements. They don't work outside the full Agricraft source. BLOCKER!

(3) Any good(!) reason for ItemWithMeta? That's what an ItemStack is...

InfinityRaider:
1) I'm aware of that, my apologies will be fixed when I complete the PR
2) That's what I meant when I asked about compiled code in the API, so okay I can follow you in that.
3)  I'm not sure, Imasius added that in for something, don't remember what exactly he just needed it for. I guess he didnt need the NBT and stacksize I suppose.

HenryLoenwind:
(1) No problem, I'm able to add dependencies on my own. As long as I don't derp it *g*

(2) Ah, ok. I understood ".class" files. No, everything the classes and interfaces reference must be included in the API jar (or be provided by Java/Minecraft/FML/Forge).

(3) I can have a look at its usage. Just not using a couple of fields would be a bad reason, but maybe it was about a "useMeta" flag. In that case, I'd recomment "class x { ItemStack a; boolean matchMeta; boolean matchNBT; }"

PR #166, opened 2015-04-16
Update fr_FR.lang

Mazdallier:
Nitor wart addition.


InfinityRaider:
By the way, also check the cropRewrite branch, I added some journal entries for some more seeds from other mods.

PR #170, opened 2015-04-18
Update zh_CN.lang

3TUSK:
Somebody told me their game crashed when using this lang file... Try fixing it by adding some space.
Does the space means "start a new line"? I checked related source code and it implies such a feature.


InfinityRaider:
Well it depends on the font really it tries to make sure the sentences stay on the left page, so when a line would exceed the maximum length, it just cuts it off at the previous space and starts a new line.

PR #176, opened 2015-04-20
Make API compile outside Agricraft

HenryLoenwind:
Only very simple injection of interfaces. The content of the interfaces
should be checked, I just took all public methods.


InfinityRaider:
Is it really not possible to ship abstract classes into the API? I'm mostly talking about the CropPlant class, I mean if people create a class that extends it (which they should) can't they just have some kind of boolean making sure the classloader doesnt load that class if agricraft is not present?

HenryLoenwind:
Any class that subclasses your class would crash unless the API was present. Same goes for any class that has an object of that class as a field. One can get around that by putting another layer of indirection around it, but: People don't want to have 2 different classes for their plants (one that works always and one that can only be loaded when Agricraft is there). This base class may be of interest for mods that *require* Agricraft. But then there's no need for it to go into the API jar, those mods can depend on the main jar.

But I actually removed it because it has dependencies on the rest of Agricraft, stopping the API from being used at all. To test that, just create a new empty mod and dump the api package (the api jar's content) into it. If that doesn't compile, nobody can use the API at all.

InfinityRaider:
Okay I see what you mean

PR #193, opened 2015-04-30
Update zh_CN.lang

UUUii:
Yeah,I think it can fix the crash when using 3TUSK's lang file.


InfinityRaider:
I'm having a bit of trouble merging this, please bare with me.

UUUii:
ooo,OK,I see.

PR #222, opened 2015-06-08
Update TileEntitySeedAnalyzer.java

iarspider:
One more place where one should sanitize analyzed seed list.


InfinityRaider:
I don't see why not, thanks.

PR #348, opened 2015-09-01
Master

RlonRyan:
Some improvements to the code, removing some of the magic numbers and spelling errors. 

This PR makes weeds customizable from the configuration files, similar to other crops.


RlonRyan:
Very well, but perhaps a new api might be a good idea in the future.

RlonRyan:
Well I had the thing fixed, then git decided to dump the commit... It was more of a few suggestions anyway.

InfinityRaider:
I quickly went over your suggestions, and I agree with most of them. Thanks for your effort.
I am willing to merge almost all of them, except for the ones that change the API. Fertiliser is according to British spelling, while Fertilizer is the American one, and I prefer the British.

RlonRyan:
Well then... Let's try this again...

*Dang British... Always misspelling things.*

PR #360, opened 2015-09-08
Fix #359

### LINKED ISSUE #359: Wooden Grate are un-Hittable from certain Angles
As illustrated below.

RlonRyan:
Cleaned up the Grate classes a bit. Seems to fix the hitbox issue.


RlonRyan:
Hold off on merging this for a moment.
 I am about to perfect it!

RlonRyan:
Got it! Grates now work perfectly.

InfinityRaider:
It appears to me you just permutated the orientation id's, so nothing really changed. The only difference is that you don't call the setBounds method anymore. I've removed the call on my branch and it indeed was fixed.
Also, your branch is one commit behind.

PR #361, opened 2015-09-09
Fix #359

### LINKED ISSUE #359: Wooden Grate are un-Hittable from certain Angles

RlonRyan:
Cleaned up the Grate classes a bit. Seems to fix the hitbox issue.

Perfected!


InfinityRaider:
Thanks

PR #363, opened 2015-09-09
Massive TileEntity Update

RlonRyan:
Cleaned up the internal TileEntity hierarchy. Should make the code more
efficient and readable. Furthermore, implemented some class modifiers to
force proper usage.


src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityCustomWood.java
```diff
@@ -13,112 +14,183 @@
...
+    private int materialMeta = 0;
 
-    //this saves the data on the tile entity
     @Override
     public void writeToNBT(NBTTagCompound tag) {
-        if(this.material != null) {
-            tag.setString(Names.NBT.material, Block.blockRegistry.getNameForObject(this.material));
-            tag.setInteger(Names.NBT.materialMeta, this.materialMeta);
-        }
+    	//Can never be null. (Has to be made out of something... lol).
```
InfinityRaider: Can be null if someone removes a mod which added a certain wood type.

src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntitySeedAnalyzer.java
```diff
@@ -33,14 +50,15 @@ public void writeToNBT(NBTTagCompound tag) {
...
-    //this loads the saved data for the tile entity
     @Override
     public void readFromNBT(NBTTagCompound tag) {
+    	//Mandatory call to super().
         super.readFromNBT(tag);
         if(tag.hasKey(Names.Objects.seed)) {
             this.specimen = ItemStack.loadItemStackFromNBT(tag.getCompoundTag(Names.Objects.seed));
         }
         else {
+        	//Not certain this is required... Unsure if networking thing?
```
InfinityRaider: Yes, if the stack is taken out, then it should also be removed clientside.
Thus if there is no tag present, there should not be a stack in that slot.

src/main/java/com/InfinityRaider/AgriCraft/tileentity/irrigation/TileEntityValve.java
```diff
@@ -53,6 +53,6 @@ public void addDebugInfo(List<String> list) {
         list.add("  - State: "+(this.isPowered()?"closed":"open"));
         list.add("  - FluidLevel: " + this.getFluidLevel() + "/" + Constants.BUCKET_mB / 2);
         list.add("  - FluidHeight: " + this.getFluidHeight());
-        list.add("this material is: " + this.getMaterialName() + ":" + this.getMaterialStack().getItemDamage());
+        list.add("this material is: " + this.getMaterialName() + ":" + this.getMaterialMeta()); //Much Nicer.
```
InfinityRaider: Agreed.

InfinityRaider:
I'll merge this tomorrow

src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityCustomWood.java
RlonRyan: That's for the right though, and I am pretty sure that the way I have the class set up, this.material cannot be null, as all the setters default to the planks.

InfinityRaider: I'm going to change it to use the getters so it'll go to default if it's null anyway,  just in case.

RlonRyan:
How's this work for a temp SeedStorage Controller?

src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityCustomWood.java
RlonRyan: The getters don't have null checks in them either. They aren't needed. But I agree with that it is better to be safe than sorry, so...

The only worry is the number of method calls.

RlonRyan:
Well check this out:

InfinityRaider:
I don't like that model. It's too similar to the regular ones. I have a couple of ideas floating around in my head about how I want to do it, but I'm not 100% sure yet.

InfinityRaider:
I'm fine with everything except the changes concerning CreativeTabs, tooltips and the new model.

RlonRyan:
See the tooltips work:

InfinityRaider:
That's not the issue, there should be an additional tooltip saying something about applying vines.
Ditto with the Valves, there's a tooltip saying they're toggled by a redstone signal.

RlonRyan:
Got it.

RlonRyan:
While we are at it, since the sprinkler is one item in the inventory, I recommend that we make an icon for it, instead of rendering it in the inventory.

PR #364, opened 2015-09-10
Huge Update

RlonRyan:
As per #363, but with the fixes wanted.


RlonRyan:
I hope this looks good to you!

RlonRyan:
What would you think about a connector channel?
Or what about trying to set the tank bounds based on the state?

InfinityRaider:
Haven't merged it due to lack of time yesterday.
And what do you mean with setting the tank bounds based on the state?

RlonRyan:
I mean somehow possibly setting the tank bounds so you could fall in? It might be a little too involved though...

RlonRyan:
When do you think the next snapshot should come out? Its been almost a month since the last beta, and it might be a good idea to get some testers on the progress...

You could probably disable the seed storage for a single commit, and do a release...

InfinityRaider:
Seed Storage Controllers will be delayed for a while, I just want to finish the Peripheral GUI and rendering before I do another release.

RlonRyan:
Sounds good. Are we a go with the merge for now?

PR #370, opened 2015-09-14
Removed Some More Magic Numbers

RlonRyan:
Nothing really big this time round... In theory we could get to the
point that in one line you could set the relative sizes of all the
blocks, so long as we centralize these constants.

### DIFF
--- a/src/main/java/com/InfinityRaider/AgriCraft/tileentity/irrigation/TileEntityChannel.java
+++ b/src/main/java/com/InfinityRaider/AgriCraft/tileentity/irrigation/TileEntityChannel.java
@@ -8,18 +8,32 @@ import com.InfinityRaider.AgriCraft.tileentity.TileEntityCustomWood;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.StatCollector;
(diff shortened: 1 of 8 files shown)


src/main/java/com/InfinityRaider/AgriCraft/tileentity/irrigation/TileEntityChannel.java
```diff
@@ -8,18 +8,32 @@
...
+    protected static final int HEIGHT = MAX - MIN;
     protected static final int DISCRETE_MAX = 16;
-    protected static final float DISCRETE_FACTOR = (float) DISCRETE_MAX / 500.0F;
-    protected static final float SCALE_FACTOR = 500.0F / (float) DISCRETE_MAX;
+    protected static final int ABSOLUTE_MAX = Constants.BUCKET_mB / 2;
+    protected static final float DISCRETE_FACTOR = (float) DISCRETE_MAX / (float)ABSOLUTE_MAX;
+    protected static final float SCALE_FACTOR = (float)ABSOLUTE_MAX / (float) DISCRETE_MAX;
 
     private int lvl;
+    private boolean multiConnect = false;
```
InfinityRaider: What's this for?

PR #381, opened 2015-09-18
Rendering Changes.

RlonRyan:
Changed channels over from the old position system to using ForgeDirections. Also cleaned up imports to remove .* imports. I was unable to fix the odd lighting on channels though...


RlonRyan:
Oh, and I moved some file stuff over to Java 7, as it is cleaner.

InfinityRaider:
Shuffling all those imports around makes it really hard for me to see what actually changed.

InfinityRaider:
And why did you get rid of those color multipliers? I spent quite some time addign those in.

src/main/java/com/InfinityRaider/AgriCraft/tileentity/irrigation/TileEntityChannel.java
```diff
@@ -118,25 +117,27 @@ public final void findNeighbours() {
...
+        	return (tileEntityAt instanceof IIrrigationComponent) && (this.isSameMaterial((TileEntityCustomWood) tileEntityAt));
         }
-        return (tileEntityAt!=null) && (tileEntityAt instanceof IIrrigationComponent)  && (this.isSameMaterial((TileEntityCustomWood) tileEntityAt));
+        return false;
     }
     
-    public boolean hasNeighbour(ForgeDirection direction) {
-        int ordinal = direction.ordinal() - FORGE_DIRECTION_OFFSET;
-        return ordinal>=0 & ordinal<neighbours.length && neighbours[ordinal]!=null;
+    public IIrrigationComponent getNeighbor(ForgeDirection direction) {
```
InfinityRaider: This method was different for the server and the client. On the server the neighbours get checked every tick, so it is better to cache them. On the client they only get checked when the chunk's vertex buffer is reloaded, so it's not necessary to cache them.

PR #385, opened 2015-09-21
Separate MultiBlocks from Tanks

RlonRyan:
Hopefully this will make the tanks more manageable.

### DIFF
--- a/src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityMultiBlock.java
+++ b/src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityMultiBlock.java
@@ -1,0 +1,375 @@
+
+package com.InfinityRaider.AgriCraft.tileentity;
+
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


RlonRyan:
Man this is quite the task...

RlonRyan:
This multi-block stuff doesn't seem to want to behave as it should...

RlonRyan:
The problem is that I would like all the multiblocks to point to a single controller, but there is no good way to save the pointer to that controller, meaning that the controller would have to be reconstructed on load, an inefficient proposition.

InfinityRaider:
Can you make this PR on the tank branch please?

RlonRyan:
Will do.

RlonRyan:
Yeah, this does need some co-op work.

RlonRyan:
Sorry for taking so long just finishing up a few more tweaks...

RlonRyan:
Ok. About to commit... The only thing I can't perfect is the multiblock breaking as it seems to get called too many times.

PR #387, opened 2015-09-22
Master

RlonRyan:


RlonRyan:
I see why other mods always have you click on the structure with some kind of tool. The onPlace() function is pretty useless as it is unreliable. I'm thinking our best option is having users right click with a hoe?

InfinityRaider:
No, its working fine now, so Its possible to make it work without a tool.

RlonRyan:
We can see. It's just the breaking and reforming can happen multiple times in the current implementation, leading to things like the created water issue.

RlonRyan:
Wait do I get commit access to the branch, or do you want me to keep creating pull requests?

InfinityRaider:
Can I give you commit access to one branch?

RlonRyan:
I haven't the slightest idea. It seems like that should be a feature, although it probably isn't.

I'll keep making pull requests.

RlonRyan:
I think I finally found why the multiblock tanks weren't responding to my breakMultiBlock() changes... the tank class overrode the method... lol!

RlonRyan:
Ok. Just down to figuring out how to get the multiblock to form when a block is placed.

RlonRyan:
This time I really think that I am down to one problem: How to set the destroy particle texture.

Any ideas?

Issue #382, opened 2015-09-18
Water duplication bug


InfinityRaider:
Placing an empty water tank next to one holding water duplicates liquid.

RlonRyan:
Yeah I noticed some fluid duplication... I think part of the issue is all the casts...

In truth, no offense, the whole irrigation is kinda a mess...

InfinityRaider:
I know, it's still code from the initial alpha days, where I didn't know what I was doing.

RlonRyan:
I know the feeling.

(I might have tried to redo parts of it... but it's hard.)

RlonRyan:
By the way, are the increasingly disparate fluid level syncs on larger and larger tanks intentional? As in a single tank will sync every few hundred millibuckets, whereas larger ones will only sync in after changes of multiple buckets?

InfinityRaider:
Well, the height difference remains the same, the volume of water increases, but visually it doesnt, its separated in 32 discrete levels.

RlonRyan:
It doesn't as the 32 discrete levels grow increasingly far apart, as they are a function of the tank's maximum capacity.

RlonRyan:
I mean you don't want the syncs to occur too frequently that it bogs down the server, but they shouldn't be so disparate that it makes it look like the water level isn't changing to the client.

InfinityRaider:
Ye but if the discrete level wasnt there, the level would also rise slow/fast depending on the footprint of the tanks.

RlonRyan:
I know that. I mean that it will show as the same fluid level in WAILA for longer periods of time, to the point that clicking on a tank with a water bucket looks like nothing was added.

RlonRyan:
Perhaps implement a max sync delta at 1 bucket?

Darquewolf:
just as a note on this bug it was not in version 1.3.1 but in the new 1.4.0 its there.

https://youtu.be/HOLT__aWQvk

InfinityRaider:
We are aware, its a consequence of a major performance improvement.

InfinityRaider:
Fixed by #404

Issue #312, opened 2015-08-04
Fix tanks crossing chunk boundaries bugging out on chunk loading


Terpo:
Happens for our map on server and clientside. Forge asks for missing blocks/items, creates the backup and tries to load the map. While loading the whole game stalles. Client is sitting in the MC main screen and trying to connect to a server (with stalled map) causes a simple timeout.

Forge 1448
AgriCraft-1.7.10-1.4.0-beta-1-1.7.10.jar

http://blog.terpo.org/upload/mc/modlist.txt

http://pastebin.com/fuqY2mMV
http://pastebin.com/NHAUuXfL

It only happens for our large server map. A small test world is loading fine. Could this happen if a large tank is not fully chunkloaded? After removing one large tank the map is loading, but it stalls again with the same issue.

InfinityRaider:
What version did you update from?

InfinityRaider:
Apparently I was saving the block name instead of the Block object itself, dont ask me why, I don't even know myself. I changed that, itll improve a little, but I don't think itll fix this.

If you can start the server on an older version, I recommend changing all tanks so that they don't cross chunk boundaries.

InfinityRaider:
Another thing that crossed my mind, did you remove any mods that add planks that might have been used in the creation of a tank?

Terpo:
I told my member today that everybody should respect the chunk boundries and I will also give it a last test tomorrow. After deleting the tanks in SP (testwise) it did work. So it should be ok for the update hopefully :). 

The server update is from AgriCraft-1.7.10-1.3.1

InfinityRaider:
Well I'm going to have to something about this, so I'll leave this open.

InfinityRaider:
haven't been able to reproduce after #404

PR #388, opened 2015-09-23
Tank Fixes

RlonRyan:
Fixes Tanks.

Just have to fix the manual addition stuff.

### DIFF
--- a/src/main/java/com/InfinityRaider/AgriCraft/items/blocks/ItemBlockCustomWood.java
+++ b/src/main/java/com/InfinityRaider/AgriCraft/items/blocks/ItemBlockCustomWood.java
@@ -39,27 +39,6 @@ public class ItemBlockCustomWood extends ItemBlockAgricraft {
     }
 
     /**
-     * Places the associated block at a location in the world.
-     * There is generally no need for overriding this method in subclasses.
-     * If the block can only be placed in certain spots, override the canPlaceBlockAt() method in the block class.
-     */
-    @Override
-    public boolean placeBlockAt(ItemStack stack, EntityPlayer player, World world, int x, int y, int z, int side, float hitX, float hitY, float hitZ, int metadata) {
-        if (!world.setBlock(x, y, z, field_150939_a, metadata, 3)) {
-            return false;
-        }
-        if (world.getBlock(x, y, z) == field_150939_a) {
-            field_150939_a.onBlockPlacedBy(world, x, y, z, player, stack);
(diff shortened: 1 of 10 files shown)


RlonRyan:
Should close #382.

src/main/java/com/InfinityRaider/AgriCraft/items/blocks/ItemBlockCustomWood.java
```diff
@@ -39,27 +39,6 @@ public ItemBlockCustomWood(Block block) {
     }
 
     /**
-     * Places the associated block at a location in the world.
```
RlonRyan: This makes it so there is no weird flash of the default material before the actual material shows up.

src/main/java/com/InfinityRaider/AgriCraft/tileentity/TileEntityMultiBlock.java
```diff
@@ -13,11 +13,7 @@
 
 public abstract class TileEntityMultiBlock extends TileEntityCustomWood implements IDebuggable {
 
-	/**
-	 * Represents if the multiblock is pre-1.4
-	 */
-	private boolean oldVersion = false;
-
+	private boolean reform = false;
```
RlonRyan: This is the same thing as the oldVersion boolean, just put to more use.

RlonRyan:
Oh the pauses in the water levels reflects the delay in syncs... Perhaps the discrete level syncing isn't fast enough? I might need to adjust the levels...

RlonRyan:
Syncing every 250mB sounds o.k.? (Not to frequent, not to infrequent?)

RlonRyan:
Might also affect #312.

RlonRyan:
Oops. You might want to check the onBlockActivated() method, and somehow get it to just ignore water applications and not everything... :smile:

RlonRyan:
Hmm... I think I am going to change that mass of variables, getters, and setters at the top of the tileentityMultiBlock file to an immutable object.

RlonRyan:
I think I might just have spent an hour that was due to an NBTTag spelling error...

RlonRyan:
I think this works pretty nicely.

RlonRyan:
I'd say that this is a go on the merge.

InfinityRaider:
I'm gna do some testing, but I'm going to wait with merging this into master until after the next release.

RlonRyan:
Ok.

RlonRyan:
Wait, why are we holding off the update?

PR #390, opened 2015-09-25
Irrigation Tweaks

RlonRyan:
Just a few minor things. Managed to fix the controller component being
reformed every fluid addition.

I think this should make the branch all ready to merge back into master.


InfinityRaider:
Thanks, but I'm still going to wait with merging into master, reason being that this is a big code change to a core mechanic and I'd like to test that myself extensively first, as well as put out beta builds with this.
I don't have a lot of time lately and I'm so close to the next release that I don't want to delay it any more.

RlonRyan:
I hear you, and thoroughly agree with your reasoning.

PR #392, opened 2015-09-27
Fast Gradle

RlonRyan:
Try to speed up Travis, and clean up Gradle.


RlonRyan:
That first run will be slow because the cache hasn't been generated yet.

InfinityRaider:
Thanks

PR #397, opened 2015-09-29
Merged Master Into Tank

RlonRyan:
So as to gain the benefits of the recent additions to master.


RlonRyan:
Nope... didn't merge this right... Looks like I'll have to try again.

PR #398, opened 2015-09-29
Merge Master Into Tank

RlonRyan:
Got the right branch this time, directly from the main repo.

Sorry for the dupe.


RlonRyan:
On second thought, it might be best if you were to do this directly, since this creates extra merge commits.

RlonRyan:
Shoot, I guess the source was right the last time, so all I did was make a duped pr... Sorry!

PR #409, opened 2015-10-06
Added Mutator Item

RlonRyan:
So seeds may be mutated in creative quickly.


RlonRyan:
Just needs 2 more things:

1. Combining recipes. (Not crafting recipes).
2. Textures... (Beyond my skill).

RlonRyan:
Well adding the recipes might be hard, as there are (3^3) * (n^3) possible items...

InfinityRaider:
Thanks, but I deliberatly never added something like this to avoid cheating.

RlonRyan:
I figured as much. Still, it was good practice for me, so not a total waste.

InfinityRaider:
I'll merge it and play around with it, I'm going to hide it from NEI though, or I might only register it if a dev env is detected.

RlonRyan:
Ok. Cool. Yeah, I agree to keep it away from non-creative as it would break the gameplay. It was more of a tool for the server admins who wanted to replace things that got lost for some reason.

RlonRyan:
I think using the generic crop implementation, we might be able to cut out half of the api classes, and have the same functionality.

RlonRyan:
Durp. Wrong java version number.

InfinityRaider:
Can you make a different PR for this, and Unchange allt he changes to API methods?

RlonRyan:
I'll make a personal branch for the last two commits then.

InfinityRaider:
Can you make two different PR's when you find the time? One for the mutator and one for the formatting (without the change to the API methods).

PR #433, opened 2015-10-26
Use game profile name in packets instead of display name

sk89q:
This changes MessageAgriCraft to use the player's GameProfile name instead of the display name, as the display name on the server could differ from the one on the client.


InfinityRaider:
Thanks

PR #465, opened 2015-11-16
Chinese Update Part 1

claybiockiller:


InfinityRaider:
To be continued?

claybiockiller:
Yes, it will be continued later.

PR #466, opened 2015-11-17
Chinese Localization Part 1.5

claybiockiller:


3TUSK:
Seriously, but why not squash into _only one_ commit?

claybiockiller:
3T is everywhere......
Because I sometimes stop translation of Agricraft and check out related mod......
3T无处不在。。。。。。
因为我有时会暂时中断农业工艺汉化并查看联动模组。。。。。。
Claybiockiller(Github)/Clayblockiller(MCF)/Clayblock(MCBBS)

liach:
@3TUSK Most localization contributors edits online and do not squash commit as they do not have a client.

InfinityRaider:
Although single commits are cleaner I do not really mind, so it's fine guys.

3TUSK:
@liach I usually either copy the `en_US.lang` to local or execute `git clone` when I lose my mind, since I know I would pause somewhere. I always try making everything clean.

PR #475, opened 2015-11-23
Fixed Shovel Interaction Bug

oitsjustjose:
To go into further detail: Many people complain that when clicking on chests / crafting tables / doors / etc.., the swing is nullified, and sometimes other things too (like opening a GUI, the door open sound, etc.). This proposal changes flag to false if the block clicked is NOT a farmland block, preventing any other code that may result in this weirdness from occurring.


InfinityRaider:
Oh sorry about that, also I'm going to move this to the very start of the method, before checking equipped items and remove the check further down. Thanks for this PR.

PR #481, opened 2015-12-02
Add ability for precise mod-specific configuration

keepinggiant:
I was looking through the mod description when I saw a line that you could disable the mutation recipes in NEI, but I couldn't find any configuration options to do so. I decided to add that functionality. At the same time, I made it easier to add similar configuration options in the future.


InfinityRaider:
Hmm, that config option seems to have been lost a while ago when I changed my compat system.
Thanks for your effort, but I don't really like the 'enableCompatibilityConfiguration(String modId)' method.

Also I'm going to redo this, because with this players can jsut toggle the value client-side, I'm going to implement a config option which enforces this config setting from the server.

PR #499, opened 2015-12-14
Preserve compatibility with APIv1 plants

HenryLoenwind:


HenryLoenwind:
I'm not happy with the new getter in CropPlantHandler, but keeping a second list in APIimplv1 would mean updating a fully registered plant was possible. That I wanted to avoid.

InfinityRaider:
This is clever, I like it. I think I'll undeprecate the registerGrowthRequirement() method and repurpose it as a method to override growth requirements. This will also get rid of that getter.

PR #555, opened 2016-02-07
Continued Updates

RlonRyan:
Attempted to improve code more.


RlonRyan:
I still have no idea why the CustomWood blocks are suddenly undifferentiated in 1.8...

RlonRyan:
The other thing is, if we ever wanted to rewrite the api, this would be the time.

RlonRyan:
Should we completely remove the travis automation stuff?

RlonRyan:
Also, why isn't the AgriCraft icon on CurseForge the normal one?
It kinda bugs me...

InfinityRaider:
Idk, what do you mean with AgriCraft icon?
Also what is up with all those reordering of imports?

RlonRyan:
The Wheat and Hoe icon... (At least I was under the impression that was the agricraft icon, it is after all on the creative tab and mod info page...)

I can undo the import reorder commit, I was just trying to remove all the unused imports and clean up a bit...

InfinityRaider:
I don't mind that, the only proble is that when I run an optimize imports it'll shuffle it again, and then when you do it again it'll be shuffled again and it'll uglify the commits.

RlonRyan:
Ah. I see, want me to try to remove the commit then?

InfinityRaider:
Would be appreciated.

RlonRyan:
Will do.

RlonRyan:
I still think that an optimize imports is needed. It seemed to remove some build hiccups...

RlonRyan:
We need to switch the root classes to interfaces...

InfinityRaider:
Why exactly?

RlonRyan:
Ok, maybye not...

RlonRyan:
It's just that the seeds, which are items do not extend the agricraft item class...

RlonRyan:
Btw: I updated the icon... :hint:

InfinityRaider:
I did that deliberately to keep the vanilla ItemSeeds implementation, in case some mods check for ItemSeeds. About the icon, I know I saw, but I don't think some mods use the same naming for their items and icons like vanilla does.

RlonRyan:
Yeah I figured we needed to keep the seeds subclasses of ItemSeeds, it just complicates the internal class structure, which is why I thought an interface might work better. However, it might be better for us not to consider seeds items, and rather a subclass of ItemSeeds...

RlonRyan:
Oh and the Icon was the .png for the CurseForge... I don't know what you are talking about for the naming though, are you referring to SafeIcon?

InfinityRaider:
Yes

RlonRyan:
Yeah that's more for internal loading, and might be temporary...

PR #561, opened 2016-02-08
Item Rendering Hack

RlonRyan:
Got some more things to render.


RlonRyan:
Well all the 1.8 JSON item tutorials I've tried don't work... so there's that...

TehNut:
What have you been trying? Maybe I could help

RlonRyan:
Do you know of any simple example mods that actually compile? I want to see if there is a way to switch over to JSON or not...

TehNut:
You *should* just be able to register your item, then do

`ModelLoader.setCustomModelResourceLocation(YourItem, META, new ModelResourceLocation(new ResourceLocation(YOURDOMAIN, REGISTEREDNAMEOFITEM), "variant=type"));`

Just that will let you do all the Item model stuff with a single JSON in `assets/modid/blockstates` with a filename of `REGISTEREDNAMEOFITEM.json`

[This](https://gist.github.com/TehNut/a72f835ba36a995851c9) is an example that *should* work. Completely pseudocode, though.

RlonRyan:
I'll have to try that, thanks!

InfinityRaider:
There's plenty of 1.8 examples, however not for things like the Custom Wood stuff, there a model is rendered with a texture based on the NBT tag of the ItemStack, and I have no idea how to do that with this json thing.

InfinityRaider:
Also why did you get rid of the render class for the clipping? It doesn't just render an icon.

src/main/java/com/InfinityRaider/AgriCraft/proxy/ClientProxy.java
```diff
@@ -78,14 +80,24 @@ public void registerRenderers() {
             if(field.getType().isAssignableFrom(ItemBase.class)) {
                 try {
                     Object obj = field.get(null);
-                    if(obj!=null) {
-                        ((ItemBase) obj).getItemRenderer();
+					// This is a safer check
```
InfinityRaider: If obj is null, obj instanceof ItemBase will be true, and this will NPE.

RlonRyan: http://stackoverflow.com/questions/2950319/is-null-check-needed-before-calling-instanceof

RlonRyan:
I got rid of it because it didn't do anything at the moment. It can always be re-added when the core stabilizes more.

src/main/java/com/InfinityRaider/AgriCraft/proxy/ClientProxy.java
InfinityRaider: I'm 100% sure I did that before and I got NPE's, not here but somewhere else.

InfinityRaider:
It does something, it adds a render of the mature plant on top of the seed bag.

RlonRyan:
Yeah I'll add it back in.

RlonRyan:
I just can't tell if the JSON rendering would be better or not...

RlonRyan:
Ok, I've started to get a [handle](https://github.com/RlonRyan/ExampleMod) on 1.8... I'll add the clipping renderer back shortly.

RlonRyan:
Ok, that should do it! :shipit:

src/main/java/com/InfinityRaider/AgriCraft/proxy/ClientProxy.java
RlonRyan: Yikes.

RlonRyan:
Then just pull this for now, and we can move it later.

InfinityRaider:
Why not pull it yourself?

RlonRyan:
Ok then...

PR #567, opened 2016-02-15
Correct Crop Rendering

RlonRyan:
And with that, the only thing left to get to render are the custom wood
blocks in the inventory! :yay:

Oh, and I think I found the missing ghost import...

### DIFF
--- a/src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockBase.java
+++ b/src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockBase.java
@@ -113,4 +113,9 @@ public abstract class BlockBase extends Block implements IAgriCraftRenderable {
 		this.icon = iconRegistrar.registerIcon("agricraft:blocks/" + name);
 	}
 
+	@Override
+	public EnumWorldBlockLayer getBlockLayer() {
+		return EnumWorldBlockLayer.CUTOUT;
+	}
+
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockBase.java
```diff
@@ -113,4 +113,9 @@ public void registerIcons(IIconRegistrar iconRegistrar) {
 		this.icon = iconRegistrar.registerIcon("agricraft:blocks/" + name);
 	}
 
+	@Override
+	public EnumWorldBlockLayer getBlockLayer() {
+		return EnumWorldBlockLayer.CUTOUT;
```
InfinityRaider: Maybe return the Default value you have set here instead?

RlonRyan:
Ooops... Sorry...

src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockBase.java
RlonRyan: What do you mean?

RlonRyan: I just added back in the method I shouldn't have deleted...

InfinityRaider: Its only minor, but return this default field: https://github.com/InfinityRaider/AgriCraft/blob/master/src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockBase.java#L25

RlonRyan: Will do!

PR #570, opened 2016-02-16
Name Refactoring

RlonRyan:
The goal of this branch is to remove the Names class, as to clean up the code.


RlonRyan:
I guess the issue with the running PR is that the PR fails if the main branch is failing... but the main branch should probably never be allowed to be failing.

RlonRyan:
The way I have Travis setup, we should keep main the working branch, from which we release, and we work in dev- branches, if that is good with you.

RlonRyan:
I think that should help some...

src/main/java/com/InfinityRaider/AgriCraft/utility/RegisterHelper.java
```diff
@@ -34,15 +38,34 @@ public static void registerBlock(Block block, String name, Class<? extends ItemB
...
-                if (fruit != null && fruit.getItem() != null && !OreDictHelper.hasOreId(fruit, Names.Objects.crop + name)) {
-                    OreDictionary.registerOre(Names.Objects.crop + name, fruit);
+                if (fruit != null && fruit.getItem() != null && !OreDictHelper.hasOreId(fruit, "crop" + name)) {
+                    OreDictionary.registerOre("crop" + name, fruit);
                 }
             }
         }
     }
+	
+	public static void registerItemRenderer(Item item) {
```
InfinityRaider: This will need a SideOnly(Side.CLIENT) annotation

RlonRyan: Ok

RlonRyan:
And I think that this way you could theoretically commit to the branches I work on, for edits such as that...

InfinityRaider:
Ah yes, I didn't notice it was from the dev branch my apologies.

RlonRyan:
Should I try a merge?

InfinityRaider:
Go for it.

RlonRyan:
It seems to have worked well!

InfinityRaider:
In the future, if we want to do something new, just create a new branch and make a PR for it. This is the most useful and clear imo.

RlonRyan:
So you like this setup?

PR #573, opened 2016-02-17
Refactor Patch

RlonRyan:
The goal of this branch is to refactor all the AgriCraft code to match both MineCraft and Java standards.


RlonRyan:
Ok... I think that time I got it done right!

RlonRyan:
@InfinityRaider I did it! I found the line causing the custom wood blocks to act funny!

InfinityRaider:
What line was it?

RlonRyan:
The tile entity dummy was not reloading the icon every time the material changed.

InfinityRaider:
ooooooh, good find.

RlonRyan:
So do you think the package rename will be ok?

InfinityRaider:
Its the first release, so sure.

RlonRyan:
Cool.

RlonRyan:
That last commit was quite the upgrade, if I do say so myself.

RlonRyan:
So... I think this branch is good to go... I kinda got off track on the renderers...

Renderer work next.

InfinityRaider:
I don't really have any comments/remarks for this so you're free to merge it, unless you want to add some more commits.

RlonRyan:
Ok... sounds good...

Isn't it like 3AM for you?

PR #576, opened 2016-02-20
Render Branch

RlonRyan:
The goal of this branch is to simplify the rendering process.


RlonRyan:
Redoing the crop renderer is going to require some serious refactoring.

RlonRyan:
Other than that, the rendering seems pretty good now...

RlonRyan:
I might need your help with hacking in the dropped item rendering...

InfinityRaider:
What do you mean with dropped item rendering? doesn't it work anymore? It used to work.

RlonRyan:
It works with the the items that use JSON, just not our custom items.
They just render as the default texture block.

InfinityRaider:
Hmm, it worked in 1.8.8, I'll have a look at it.

RlonRyan:
Ok, thanks.

RlonRyan:
I don't know what you think about the state of this branch, but I think it might be good to go for a merge.

InfinityRaider:
I'm alright with it, so you have a green light from me.

RlonRyan:
Sounds good!

PR #582, opened 2016-03-04
Continued Render Updates

RlonRyan:
This is a continuation of the render branch, which was previously merged as to preserve progress.


RlonRyan:
I plan to remove the item rendering classes completely, now that we have an alternate way of rendering items... That is, so long as it is ok with you.

InfinityRaider:
If it works, and it doesn't use that hack I'm totally fine with it.

RlonRyan:
I just have to figure out how to correct the held item orientation...

RlonRyan:
Well... I got it down to only having 60 model load failures...

RlonRyan:
lol... WAILA for 1.8.9 still works on our blocks, but not a vanilla lever...

RlonRyan:
Well, this is a dev branch after all, so we keep working on it until it gets to a state at which it is appropriate to merge.

RlonRyan:
We should probably combine the seeds like the clippings...

Also, is there any reason that AgriCraftCrops have to be blocks? They are never used as actual blocks...

RlonRyan:
Anyways, the clipping rendering seems pretty good now, although the plant icons on them are a *bit* too big... lol.

InfinityRaider:
They can all be grown outside of crop sticks, idk they started of like that when the code was based on vanilla's implementation of crops and seeds. But then I'd have to start writing exceptions for some mods which don't follow vanilla's implementation. At a certain point there were so many exeptions that I decided to make my own interface for it and base everything on that.

RlonRyan:
Well I guess if they need to be grown outside of crops that is fine... but not having them as blocks might simplify a lot of things...

InfinityRaider:
Why is that? They are identical to wheat, just with other textures, so we can just copy the json for wheat and modify the texture name.

RlonRyan:
As in we don't have to worry about registering blocks or loading JSON files for them...
And to have everything done through an interface would be pretty nice...

InfinityRaider:
I don't understand the json thing yet, but when I do I'm sure itd be possible with Forge to have a single json and just use the appropriate textures. The textures are named very consistently anyways.

RlonRyan:
Well that should cover any normal item models we need to generate...

RlonRyan:
I think that this branch should be good for now.

PR #585, opened 2016-03-11
Added Annotation-Based Configuration

RlonRyan:
It allows for the placement of configurable values in the class that
they are actually used. It also allows for the cutting down on the
massive ConfigurationHandler class.


RlonRyan:
I got a bit off-track on the render branch... lol.

RlonRyan:
I didn't want to commit this to the render branch, in case if you didn't like it.

Personally, I think it's quite neat.

RlonRyan:
Like I said, very far off track...

InfinityRaider:
I like the annotation based config, I was planning on rewriting it to an interface-based system to make it exposed to the api, so people could read agricraft's config settings without having to do all-too crazy things

InfinityRaider:
Just one concern, for ints the config needs a default and a range, for strings and booleans its just a default. And how do you plan to tackle arrays?

RlonRyan:
The default is the value that is already assigbed to the field, and the range is already handled with min and max.

RlonRyan:
Arrays will take a bit more work but should be doable.

InfinityRaider:
Ah yes, you just give them a range, but the values doesn't actually do anything, I see.

RlonRyan:
What values?

InfinityRaider:
for min and max.

RlonRyan:
Oh they are functional...

RlonRyan:
I think that should do it for this branch, given that you are o.k. with it.

Give me the word, and I'll merge it back into dev-renderer...

RlonRyan:
(I kinda want to add a `develop` label for these dev-branch prs...)

RlonRyan:
So...

InfinityRaider:
I couldve sworn I added that label...

InfinityRaider:
Apparently I did

Issue #593, opened 2016-03-27
[2.0.0a]Server Crash on Start


oitsjustjose:
Sorry for the vague crash-log, it seems like maybe some client-only code is being called on the server? Not a huge deal, we can wait for the awesomeness that is AgriCraft just a little while longer!

[FML Crash Log](http://puu.sh/nVdxp/dd2a4f0047.txt)

RlonRyan:
It looks like I must have missed a @SideOnly annotation somewhere...

It appears that the server side is trying to load rendering code, which is client only...

PR #596, opened 2016-03-27
Separate Client & Server Code

RlonRyan:
The purpose of this branch is to untangle the client code from the server side code.

As per #593

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/blocks/BlockModPlant.java
+++ b/src/main/java/com/infinityraider/agricraft/blocks/BlockModPlant.java
@@ -34,11 +34,13 @@ import com.infinityraider.agricraft.reference.AgriCraftProperties;
 import com.infinityraider.agricraft.utility.icon.IconUtil;
 
 public class BlockModPlant extends BlockCrops implements IAgriCraftPlant {
+	
     private IGrowthRequirement growthRequirement;
     public CropProduce products = new CropProduce();
     private ItemModSeed seed;
     public int tier;
-    @SideOnly(Side.CLIENT)
+	
+    //@SideOnly(Side.CLIENT)
     private RenderMethod renderType;
     @SideOnly(Side.CLIENT)
     private TextureAtlasSprite[] icons;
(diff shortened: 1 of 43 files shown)


RlonRyan:
This is going to take a while...

RlonRyan:
Well the server launches now... but I am not entirely sure if I should merge this...

src/main/java/com/infinityraider/agricraft/blocks/BlockModPlant.java
```diff
@@ -35,11 +34,13 @@
...
 
 public class BlockModPlant extends BlockCrops implements IAgriCraftPlant {
+	
     private IGrowthRequirement growthRequirement;
     public CropProduce products = new CropProduce();
     private ItemModSeed seed;
     public int tier;
-    @SideOnly(Side.CLIENT)
+	
+    //@SideOnly(Side.CLIENT)
```
InfinityRaider: why is this commented out?

src/main/java/com/infinityraider/agricraft/tileentity/TileEntityCustomWood.java
```diff
@@ -58,13 +59,23 @@
...
-	@Nonnull
 	@SideOnly(Side.CLIENT)
-	private TextureAtlasSprite icon = IconUtil.getDefaultIcon();
+	private TextureAtlasSprite icon;
 	
 	@Nonnull
 	private boolean isIconCached = true;
 
+	public TileEntityCustomWood() {
+		if (FMLCommonHandler.instance().getEffectiveSide().equals(Side.CLIENT)) {
```
InfinityRaider: I don't like this...

insuusvenerati:
Been running with a1 for a while on a server and it seems fine. Getting another random crash but it seems to be unrelated.

src/main/java/com/infinityraider/agricraft/blocks/BlockModPlant.java
RlonRyan: Because it caused issues...

src/main/java/com/infinityraider/agricraft/tileentity/TileEntityCustomWood.java
RlonRyan: Nor do I, but that's until I can get another method to do it.

PR #617, opened 2016-04-04
Move Classes Around

RlonRyan:
An attempt to get a better package hierarchy.


RlonRyan:
I don't know if something like this would be better in terms of structure or not... I suppose this would be better for the 1.9 branch, but at least here you can see what I am getting at.

InfinityRaider:
I don't like the client package some mods do, personal preference I suppose.

RlonRyan:
Ok!

PR #634, opened 2016-04-30
Add Dragtech Agricraft plant textures

Kupferdrache:
We want to share our agricraft crops with all so we would abrichate if you merge these new crops into the mainline.


Kupferdrache:
hm, I am curious, there are no new tags since last year, but on curse are sources as jar file labeled 1.5.0 how dow that match?

InfinityRaider:
What exactly is the purpose of this?

RlonRyan:
Seeing as there has been no reply to this PR in 20 days, I vote to close it.

PR #640, opened 2016-05-02
Long standing bug

dshadowwolf:
For a very long time AgriCraft has ignored the 'Weeds Wipe Plants'
configuration item. This is an attempt to fix that issue.

(It appears as though this was intended to be there but the code was never added)


InfinityRaider:
This appears to be a valid fix, so why are you PR'ing this 3 times and closing it again afterwards?

dshadowwolf:
Was a bug in the original couple thank to the editor I was using to view
the code (originally - I was trying to find out why "Weeds Wipe Plants"
wasn't being listened to) deciding that it knew more than me about what
went where when I made the really quick edit to fix the problem.

So I respun it and found that the CI testing was spewing still and couldn't
solve that.

DRH

PR #645, opened 2016-05-13
Fixed CropPlantAgriCraftShearable not overriding the onHarvest that i…

### LINKED ISSUE #590: [1.7.10-1.5.0] Harvesting with shears is broken.
When harvesting crops like cacti or flowers in 1.5.0 using shears, you get the same drops as when harvesting normally. Works as intended in 1.4.6.

### LINKED ISSUE #609: [BUG?] [1.7.10] Shearing cacti does not yield cactus block
I tried to shear a 10/10/10 cactus with vanilla shears and ThermalFoundation bronze shears but none yielded a cactus block, instead just harvested the crop normally (as right click do). Please restore original behavior for getting cactus block by shearing.

AgriCraft 1.5.0

Mordenkainen:
…ncludes an ICrop as a parameter. This was causing shears not to work.

Fixes #609
Fixes #590

NOTE: I re-implementsd onHarvest completely since the older one is marked as deprecated, but really the savings of avoiding a single call to world.getTileEntity almost doesn't seem worth it. I almost leaned toward just delegating the call to the existing onHarvest method. It might have been cleaner that way. Let me know if this is what you would prefer. If so, I will update the pull since I already have that code as well.

### DIFF
--- a/src/main/java/com/InfinityRaider/AgriCraft/farming/cropplant/CropPlantAgriCraftShearable.java
+++ b/src/main/java/com/InfinityRaider/AgriCraft/farming/cropplant/CropPlantAgriCraftShearable.java
@@ -7,6 +7,7 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemShears;
 import net.minecraft.item.ItemStack;
+import net.minecraft.tileentity.TileEntity;
 import net.minecraft.world.World;
 
 public class CropPlantAgriCraftShearable extends CropPlantAgriCraft {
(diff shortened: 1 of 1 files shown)


InfinityRaider:
This seems okay, I don't mind this overload at all. 
Also people tend to automate these crops. A lot of them. Any call to world.getTileEntity() which can be avoided should be.

Mordenkainen:
I do not believe an official build was ever done. You can get my unofficial build (used in the Regrowth modpack) here:

https://www.dropbox.com/d/d1tk9aeboytkgnw/AgriCraft-1.7.10-1.5.0-Unofficial.jar?dl=0

CodesCubesAndCrashes:
@Mordenkainen Thank you! I don't have 1.7.10 stuff set up, so it would have taken me longer to help with this. 
@ThuhElement Let me know how that turns out. :)

ThuhElement:
@Mordenkainen I was still able to re-create the issue using the unofficial. I used creative mode and switched to survival and used several plants and shears (I even used a silk touch mattock.) I don't think this resolved the plants dropping flowers/cactus. I raised a cacti to lvl 9 and still was not able to obtain a cactus and the same for flowers. I wish I could help more. You guys are a blessing with all the work you do either way.

@CodesCubesAndCrashes I'm sorry buddy. It still didn't work for me.

CodesCubesAndCrashes:
Hm, what plants were you trying to get flower drops from? Did they used to work at some point? I don't remember what was standard in 1.7.10 for AgriCraft, sorry. But I've looked at the call chain, and I haven't found yet any plants that are built in and that make use of the 'shearable' drop feature. I can see that it's also an feature available to the custom crops, but it doesn't look like any of those are included by default. 

Actually... could you better explain what do you mean by "get flowers from shearing" and "obtain a cactus"? I might have misunderstood before. Mm, instead do you want to make the crop drop as a block? Kind of like when you silk touch foliage on a tree or grassy dirt? Because the code in this fix is about fruit that gets harvested by using vanilla shears on the plant. And I want to make sure that that is the same thing that you are expecting to happen.

ThuhElement:
@CodesCubesAndCrashes Wow, I am so sorry. I had read on a few post from other pages before saying they were only getting dye. I assumed that this was a fix for that and I apologize terribly. I had thought people were wanting the dye "AND" the flower when they right clicked on a mature Agricraft plant. I am using Witchery now and use the actual Poppy flower and Dandelion. I must've misunderstood. Once again I apologize for the inconvenience. I know you guys do so much for the community I hate to slow you down.

CodesCubesAndCrashes:
No trouble at all! Hopefully you can make some custom plants that do the thing you want. In any case, please feel free to ask still more questions in the future. :) Cheers.

PR #646, opened 2016-05-15
Openblocks sprinklers will no longer cause growth in plants that do n…

### LINKED ISSUE #584: [Bug] OpenBlocks Sprinkler causing plants to grow when they shouldn't.
The Openblocks sprinkler causes ore plants to grow when they should not be able to grow. An example is when you're required to have the ore block under the plant, it says "It cannot grow here" but add a sprinkler and it still grows. 

This is currently AgriCraft-1.7.10-1.5.0 but my guess is it happens with all versions.

Mordenkainen:
…ot meet their growth requirements.

Fixes #584


InfinityRaider:
Cheers

PR #650, opened 2016-05-17
When emptying a water pad water containers will be ejected out into t…

Mordenkainen:
…he world if there is no room in the players inventory.

I won't put the "fixes" tag in here, since that would close the issue even though it is still open in 1.9.


InfinityRaider:
Thanks again

PR #666, opened 2016-06-19
API Overhaul

RlonRyan:
Rework API as to make it more logical and clean. There is still quite a bit of work to be done, but I would like some confirmation that I am on the right track...


src/main/java/com/infinityraider/agricraft/api/example/ExampleAgriCraftAPIimplementation.java
```diff
@@ -2,10 +2,10 @@
 
 import com.infinityraider.agricraft.api.API;
 import com.infinityraider.agricraft.api.APIBase;
-import com.infinityraider.agricraft.api.v1.APIv1;
+import com.infinityraider.agricraft.api.v3.APIv3;
```
InfinityRaider: Whats the point of v3? Its a new version of mc, might as well reset api version.

RlonRyan: That's true...

RlonRyan:
I guess I was just thinking in terms of continuity, but I like the idea of keeping the version number low too.

RlonRyan:
Also... I saw you made a repository for the API... Are we planning to move the API there?

InfinityRaider:
I was goint to do that, but I couldn't get it to work.

RlonRyan:
You want me to try and move the API for you?

InfinityRaider:
Thanks, but there is no need for it.

RlonRyan:
If you are sure.

RlonRyan:
You think I could maybye merge these soon? I fear they are getting close to being runaway branches, and I wouldn't want to waste work on something that won't even be used...

InfinityRaider:
Sure, you can merge it.

PR #671, opened 2016-06-28
Add Custom JSON Items

RlonRyan:
For the special crop drops, like resource crop nuggets.


InfinityRaider:
I don't think this should be in agricraft...

RlonRyan:
That's why I made a pr. I'll remove this branch once I trendier the AgriCore improvements to master, leaving out the custom items. Plus the whole custom item things took less than an hour to make, so no hard feelings at all there!

PR #675, opened 2016-07-10
Fix 1.10 Rendering

RlonRyan:
This PR implements a basic fix for the rendering system in 1.10.

Started to convert the old Techne models into .obj files. Having some trouble getting the rotation functionality back, however... The blockstate.json files do not seem to want to cooperate with me...


src/main/java/com/infinityraider/agricraft/utility/AxisPosition.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @author RlonRyan
+ */
+public enum AxisPosition implements IStringSerializable {
+	
+	X_NEG, X_MID, X_POS,
+	Y_NEG, Y_MID, Y_POS,
+	Z_NEG, Z_MID, Z_POS;
+	
+	static AxisPosition[][] resolved = new AxisPosition[][]{
```
InfinityRaider: Should maybe put this in a field instead of making a new array every time.

RlonRyan: What do you mean? It's a static array? It's only made once...

InfinityRaider: wow wait what, you're right, apologies.

PR #682, opened 2016-07-26
Suggested Edits

RlonRyan:


src/main/java/com/infinityraider/agricraft/renderers/blocks/RenderSeedAnalyzer.java
```diff
@@ -73,7 +75,9 @@ private void renderSeed(TileEntitySeedAnalyzer te, double x, double y, double z)
 		// Render Seed
 		if (te != null && te.hasSpecimen()) {
 			// Draw Item
+			tess.draw();
```
InfinityRaider: draw call is problematic

PR #683, opened 2016-08-01
Fixed Item Rendering

RlonRyan:
Also fixed several misc. crashes.


src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
```diff
@@ -586,11 +586,6 @@ public ModelResourceLocation getBlockModelResourceLocation() {
 	}
 
 	@Override
-	public boolean needsRenderUpdate(World world, BlockPos pos, IBlockState state, TileEntityCrop tile) {
-		return true;
-	}
```
InfinityRaider: this will be necessary for caching later

src/main/java/com/infinityraider/agricraft/compat/computer/renderers/RenderPeripheral.java
```diff
@@ -156,14 +148,19 @@ private void renderBase(ITessellator tessellator) {
...
 	@Override
-	public void renderWorldBlock(ITessellator tessellator, World world, BlockPos pos, double x, double y, double z, IBlockState state, BlockPeripheral block,
-								 @Nullable TileEntityPeripheral tile, boolean dynamicRender, float partialTick, int destroyStage) {
-		if(dynamicRender) {
-			drawSeed(tessellator, tile);
-			performAnimations(tessellator, tile, BaseIcons.DEBUG.getIcon());
-		} else {
-			this.renderBase(tessellator);
-		}
+	public void renderStaticTile(ITessellator tess, TileEntityPeripheral Tile) {
```
InfinityRaider: I still dont like this

src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
RlonRyan: It's still there, I just added it as a default method in the super interface.

PR #689, opened 2016-08-28
Dev packages

RlonRyan:
Move packages back to where I can find them.


RlonRyan:
The tiles packaged was too far nested for my tastes...

InfinityRaider:
I Don't know though, it kinda makes sense to put them in the blocks package because they are tied to the block anyway, just like the block states.

RlonRyan:
I'll just leave it for now then...

PR #698, opened 2016-09-30
Rework Mutation Engine

RlonRyan:
This commit fixes up the mutation engine so that it once again works properly.

This PR is ready for a merge!


InfinityRaider:
Does the stat calculation work the same?

src/main/java/com/infinityraider/agricraft/farming/mutation/MutationEngine.java
```diff
@@ -10,48 +10,37 @@
  * calculates the new stats (growth, gain, strength) of the new plant based on
  * the 4 neighbours.
  */
-public class MutationEngine {
+public final class MutationEngine {
```
InfinityRaider: Don't think this should be final.

src/main/java/com/infinityraider/agricraft/farming/mutation/MutationEngine.java
```diff
@@ -10,48 +10,37 @@
...
+    
+    private static final MutationEngine INSTANCE = new MutationEngine();
```
InfinityRaider: Let's abstract this out to IMutationEngine, and add it to the API, just like in 1.7.10.
Allows addons to change mutation logic.

RlonRyan:
The stat calculation follows the exact same logic.

src/main/java/com/infinityraider/agricraft/farming/mutation/MutationEngine.java
RlonRyan: Well nothing should be extending it... Especially if we extract out the interface.

src/main/java/com/infinityraider/agricraft/farming/mutation/MutationEngine.java
RlonRyan: Agreed.

InfinityRaider:
Ok good, thats what i thought, but wasn't sure.

src/main/java/com/infinityraider/agricraft/farming/mutation/MutationEngine.java
InfinityRaider: People should be extending this especially after extracting otu the interface. For example a mod could simply extend this class, override the behaviour in one case and call super() for all the other cases.

InfinityRaider: If they want a hard dependency taht is, but I think an addon mod would have a hard-dependency anyways.

RlonRyan: No since the api would hold the IMutationEngine interface, and the mod author would replace it... At least that was what I was thinking...

RlonRyan: But I guess your way would be easier, and more consistent with what we have elsewhere.

PR #704, opened 2016-10-28
Add Gradle Formatter

RlonRyan:
So that all files are kept automatically formatted.


InfinityRaider:
whats wrong with this?

RlonRyan:
You changed master, before I committed.

PR #705, opened 2016-10-28
Automate Code Formatting

RlonRyan:
Proper code formatting is now required in order for the build to pass. This is useful for requiring that contributors provide the proper code formatting in their prs.

To format the code, simply run `gradle spotlessApply`.


RlonRyan:
I have decided that automatic formatting this way is a bad idea.

PR #706, opened 2016-11-01
Update QuikMod

RlonRyan:
This commit updates QuikMod, so that this project can contain the QuikMod InfinityLib as a subproject.


RlonRyan:
Merge! Merge! Merge!

PR #707, opened 2016-11-07
Various fixes:

Mordenkainen:
- Fixed an issue where filling a Water Pad tried to add a filled liquid container to the players inventory.
- Fixed an issue where all Magic Crops tier was being set to 3.
- Fixed an issue where a crop without a plant would cause an NPE if isFertile() is called.
 was called


InfinityRaider:
I'll merge it, but a 1.7 release is very unlikely

PR #708, opened 2016-11-13
Fix crash with non vanilla wood

Volune:
This fix a crash if AgriCraft is used with another mod that add new wood planks.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockSprinkler.java
+++ b/src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockSprinkler.java
@@ -12,6 +12,7 @@ import com.infinityraider.infinitylib.block.blockstate.InfinityProperty;
 import com.infinityraider.infinitylib.block.tile.TileEntityBase;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.block.model.ModelResourceLocation;
 import net.minecraft.client.renderer.texture.TextureAtlasSprite;
 import net.minecraft.entity.player.EntityPlayer;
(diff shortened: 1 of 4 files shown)


src/main/java/com/infinityraider/agricraft/utility/CustomWoodType.java
```diff
@@ -86,8 +78,17 @@ public ItemStack getStack() {
     }
 
     @SideOnly(Side.CLIENT)
-    public ResourceLocation getTexture() {
-        return texture;
+    public TextureAtlasSprite getIcon() {
+        try {
```
InfinityRaider: put the try {} catch() inside the if(texture == null), and in the catch statement, assign the missing sprite.

src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockSprinkler.java
```diff
@@ -141,10 +140,10 @@ public boolean canPlaceBlockAt(World world, BlockPos pos) {
 	}
 
 	@SideOnly(Side.CLIENT)
-	public ResourceLocation getIcon(IBlockAccess world, BlockPos pos, IBlockState state, EnumFacing side, @Nullable TileEntityBase te) {
+	public TextureAtlasSprite getIcon(IBlockAccess world, BlockPos pos, IBlockState state, EnumFacing side, @Nullable TileEntityBase te) {
 		TileEntity channel = world.getTileEntity(pos.add(0, 1, 0));
 		if (channel != null && channel instanceof TileEntityChannel) {
-			return ((TileEntityChannel) channel).getTexture();
+			return ((TileEntityChannel) channel).getIcon();
```
InfinityRaider: Don't return null icons please, return missing sprite instead.

Volune:
Should be ok now.

By the way, I've seen a crash when breaking some of the AgriCraft blocks. Doesn't seem related with this issue.

InfinityRaider:
Probably the multiblock code, multiblock code is bugged since the 1.8+ port, I think its the last thing to finish this port.

PR #711, opened 2016-11-24
Fixup Irrigation System

RlonRyan:
This commit fixes:
- Static Sprinkler Rendering
- Black Sides on Crops & Seed Analyzers
- Rotating Sprinkler Base
- Non-Persistant Water Tanks
- Multiblock Tanks Not Working


src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockWaterChannel.java
```diff
@@ -34,14 +34,6 @@ public TileEntityChannel createNewTileEntity(World world, int meta) {
 		return new TileEntityChannel();
 	}
 
-	@Override
-	public void onNeighborChange(IBlockAccess iba, BlockPos pos, BlockPos neighbor) {
-		TileEntity te = iba.getTileEntity(pos);
-		if (te != null && te instanceof TileEntityChannel) {
-			((TileEntityChannel) te).findNeighbours();
```
InfinityRaider: why is this gone?

InfinityRaider: err, it is used to update the cached neighbours when the channel gets a block update.

RlonRyan: Ok, it just didn't seem to have any impact during my testing. I can put it back in.

InfinityRaider: It might just be a porting artefact where the super method changed signature so the override is gone, if it works fine without, I see no reason to keep it, I haven't tested the code yet, I was just questioning why it was removed.

RlonRyan:
Can this be rebased and merged now?

InfinityRaider:
Yes

PR #739, opened 2016-12-16
Fix Various Issues

RlonRyan:
As to correct things.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/api/requirement/BlockCondition.java
+++ b/src/main/java/com/infinityraider/agricraft/api/requirement/BlockCondition.java
@@ -21,6 +21,9 @@ public class BlockCondition implements ICondition {
     public BlockCondition(FuzzyStack stack, BlockRange range) {
         this.amount = stack.toStack().stackSize;
         this.volume = range.getVolume();
+        if(this.amount > 1) {
+            throw new IndexOutOfBoundsException("The required amount of blocks must be greater than zero!");
+        }
         if(this.amount > this.volume) {
             throw new IndexOutOfBoundsException("Required amount of blocks exceeds volume of range!");
         }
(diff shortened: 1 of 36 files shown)


RlonRyan:
@InfinityRaider Please review this and push agricraft-2.0.0-i0.3.0-a4 as a CurseForge alpha if you approve. We have too many downloads on an alpha that turned out to be rather non-functional...

src/main/java/com/infinityraider/agricraft/api/requirement/BlockCondition.java
```diff
@@ -21,6 +21,9 @@
     public BlockCondition(FuzzyStack stack, BlockRange range) {
         this.amount = stack.toStack().stackSize;
         this.volume = range.getVolume();
+        if(this.amount > 1) {
+            throw new IndexOutOfBoundsException("The required amount of blocks must be greater than zero!");
```
InfinityRaider: Don't like this, can be made sure in the consructor...

RlonRyan: Wait what? That is the constructor...?

InfinityRaider: I mean the FuzzyStack constructor, make sure the size is never <= 0

RlonRyan: Yeah that's a good idea.

src/main/java/com/infinityraider/agricraft/api/requirement/BlockCondition.java
```diff
@@ -21,6 +21,9 @@
...
+        }
         if(this.amount > this.volume) {
             throw new IndexOutOfBoundsException("Required amount of blocks exceeds volume of range!");
```
RlonRyan: So we catch them early.

InfinityRaider: Ok then...

Issue #829, opened 2017-02-05
[1.10.2] Seed Journal Screen Goes Black on Mouse Over


MokahTGS:
In 2.0.0-.06.0-a10 when I mouse over the seeds that I have scanned the entire screen goes black and then goes back to normal when I move my cursor away.  The journal and seeds are still visible, just the game in the background goes black.

RlonRyan:
I have this fixed in the dev-storage branch, I just haven't decided if it is stable enough as to warrant merging into the master branch.

Thanks for reporting, as it helps keep track of these issues.

PR #846, opened 2017-02-21
Upgrade GUI System

RlonRyan:
The revamping of the GUI system to work with containers is quite nice, and allows for quite the reduction in the amount of GUI code.


RlonRyan:
Closes #829

PR #847, opened 2017-02-21
Fixed mutation chance. Closes #835

### LINKED ISSUE #835: Mutations not occuring
I'm using the latest available version for 1.10.2 via curseforge. [agricraft-2.0.0-0.6.0-a10], and I've noticed that crossbreeding as well as mutation doesn't work. It seems that the strategies for spreading/mutating just aren't happening.

Update: I actually found the issue.
AgriCraft/src/main/java/com/infinityraider/agricraft/blocks/tiles/TileEntityCrop.java
Line: 81   `else if (this.isCrossCrop() && (Math.random() < AgriCraftConfig.weedSpawnChance)) {
                this.crossOver();
            }`

Naturally weedSpawnChance is low, so it's keeping mutations from occuring. No idea why spreading doesn't happen though.

**UPDATE:**  Spreading is occuring as well. So this is simply a matter of using the wrong comparison. 
`Math.random() < AgriCraftConfig.weedSpawnChance`    should be
`Math.random() >= AgricraftConfig.weedSpawnChance` in order to provide expected functionality as documented in the config.

grundyboy34:


RlonRyan:
This is not correct. Your changes would make an event with probability one never occur, instead of always occurring. The old code is correct, just the weedspawn chance is set too low.

Issue #665, opened 2016-06-18
Water tank eats glass bottles when filling tank


ryankhart:
When I am filling the wooden water tank from Agricraft using glass bottles filled with water, I expect it to empty the water out of the glass bottles and leave the empty glass bottle in my inventory. Instead, water level in the tank rises but the glass bottle is gone from my inventory.

I'm on version 1.5 of the mod. It's the latest so far on Curse.

Issue #792, opened 2017-01-14
Fonte Size Issue - 1.7.10


chaos234:
![2017-01-14_04 00 29](https://cloud.githubusercontent.com/assets/340106/21951826/1d3125de-da0e-11e6-87aa-4737ac9931b2.png)

Like the image above I can't prove any fix for #543 or better it is still such a small font which is inacabtable.
Normaly the font from the vanilla mc or from the RP should be used where the text should be defined in a language variable and then displayed with the current used font.

So please fix this since it is not readable because it is to small and no, optifine is not the problem since it is with and without it like the image above :(

EDIT: GUI Scale: Normal

Cheers

RlonRyan:
Sorry, we know that that is an issue, but given that the port to 1.10 is quite involved we have dropped 1.7.10 support to focus solely on the port. Pr's fixing issues in 1.7.10 are welcome at all times though.

PR #920, opened 2017-04-25
Fixed bug #665

liketechnik:
Added a check when filling the tank with water that checks if the player
uses a water bottle and gives back a glass bottle in that case. This check
is needed because water bottles do NOT have a container item as for example (water) buckets have which would usually lead to the 'deletion' of the bottle like shown in issue AgriCraft/AgriCraft#665.

I also realized that when removing water from the tank with stacked items (buckets & bottles) the new items do not appear in the inventory, you first have to click on them to become visible. I suppose this is because of the markDirty call on the inventory (l. 93), but I was not sure about that so I'll leave it as it is.


InfinityRaider:
Thanks

liketechnik:
No problem, I like programming, your mod and mc so I've no problem with doing some work on this wonderful mod. 

Btw I just found a fix for issue #792, but didn't have time to get a pr ready yet.

PR #960, opened 2017-05-25
Experimental API Refactoring

RlonRyan:
I am not too certain as to what I think about this... But it is a lot of work to abandon, so...


src/main/java/com/infinityraider/agricraft/api/crop/CropHarvestEvent.java
```diff
@@ -0,0 +1,39 @@
...
+    @Nonnull
+    public IAgriCrop getCrop() {
+        return crop;
+    }
+    
+    public EntityPlayer getPlayer() {
+        return player;
+    }
+    
+}
```
InfinityRaider: I like this

src/main/java/com/infinityraider/agricraft/api/crop/CropPlantedEvent.java
```diff
@@ -0,0 +1,49 @@
...
+    public AgriSeed getSeed() {
+        return seed;
+    }
+
+    @Nullable
+    public EntityPlayer getPlayer() {
+        return player;
+    }
+    
+}
```
InfinityRaider: Still good

src/main/java/com/infinityraider/agricraft/api/crop/IAdditionalCropData.java
```diff
@@ -1,18 +0,0 @@
...
-public interface IAdditionalCropData {
-
-    /**
-     * Called when the TileEntity with this data is writing to NBT
-     *
-     * @return a tag holding all needed information
-     */
-    NBTTagCompound writeToNBT();
-
-}
```
InfinityRaider: Guess noone ever used this, and the events can be used by modders to track additional data in their own way?

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
```diff
@@ -33,6 +30,13 @@
     BlockPos getPos();
 
     /**
+     * Retrieves the world that the crop is in.
+     *
+     * @return The world in which the crop is located.
+     */
+    World getWorld();
```
InfinityRaider: Useful indeed

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
```diff
@@ -54,19 +58,19 @@
...
      * @return if this crop is fertile and thus can grow
      */
     default boolean isFertile() {
-        return this.getSeed().filter(this::isFertile).isPresent();
+        return this.hasSeed() && this.isFertile(this.getSeed());
     }
 
     default boolean isFertile(AgriSeed seed) {
-        return isFertile(seed.getPlant());
+        return (seed != null) && this.isFertile(seed.getPlant());
```
InfinityRaider: All the null checks

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
```diff
@@ -84,48 +88,31 @@ default boolean spread() {
 
     Optional<IAgriSoil> getSoil();
 
-    /**
-     * Utility method to get access to the TileEntity fields and methods for the
-     * crop.
-     *
-     * @return the TileEntity implementing ICrop
-     */
-    TileEntity getTileEntity();
```
InfinityRaider: I'd keep this method, might be useful...

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
```diff
@@ -84,48 +88,31 @@ default boolean spread() {
...
+    // =========================================================================
+    // Event Methods
+    // =========================================================================
+    public MethodResult onGrowthTick();
+    
+    public MethodResult onApplyCrops(EntityPlayer player);
+    
+    public MethodResult onApplySeeds(EntityPlayer player, AgriSeed seed);
+    
+    public MethodResult onBroken(EntityPlayer player);
```
InfinityRaider: Why are these methods declared public? It's an interface...

src/main/java/com/infinityraider/agricraft/api/misc/IAgriHarvestProduct.java
```diff
@@ -0,0 +1,44 @@
...
+    
+    int getMeta();
+    
+    int getMinAmount();
+    
+    int getMaxAmount();
+    
+    int getAmount(Random rand);
+
+    double getChance();
```
InfinityRaider: I think weights (integer) would be better.

src/main/java/com/infinityraider/agricraft/api/misc/IAgriHarvestable.java
```diff
@@ -28,25 +28,6 @@
...
-
-	/**
-	 * Calls the given consumer with the assortment of all possible fruits
-	 * gained from harvesting this object. For example, to get a list of the
-	 * random fruits, pass the consumer list::add.
-	 *
-	 * @param consumer the consumer to add the fruits to.
-	 */
-	void getAllFruits(@Nonnull Consumer<ItemStack> consumer);
+	boolean onHarvest(@Nullable EntityPlayer player);
```
InfinityRaider: Not using the new event here?

src/main/java/com/infinityraider/agricraft/api/plant/IAgriPlant.java
```diff
@@ -478,9 +281,14 @@ default IAdditionalCropData readCropDataFromNBT(NBTTagCompound tag) {
     @SideOnly(Side.CLIENT)
     List<BakedQuad> getPlantQuads(IExtendedBlockState state, int growthStage, EnumFacing direction, Function<ResourceLocation, TextureAtlasSprite> textureToIcon);
 
+    // =========================================================================
+    // </editor-fold>
+    // =========================================================================
+
     @Override
-    default int compareTo(IAgriPlant plant) {
```
InfinityRaider: How are you going to sort the plants in the journal without this being Comparable?

src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
```diff
@@ -85,7 +102,70 @@ public void updateTick(World world, BlockPos pos, IBlockState state, Random rand
     @Override
     public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, EntityPlayer player, EnumHand hand,
             ItemStack heldItem, EnumFacing side, float hitX, float hitY, float hitZ) {
-        return this.getCrop(world, pos).map(crop -> crop.onCropRightClicked(player, heldItem)).orElse(false);
+
+        // Step 0. Abort if remote.
+        if (world.isRemote) {
+            // I'm not sure if this is right, but oh well.
```
InfinityRaider: Yes, no logic should run on the client, it should be on the server and then synced to all clients.

src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
```diff
@@ -85,7 +102,70 @@ public void updateTick(World world, BlockPos pos, IBlockState state, Random rand
...
+            return false;
+        }
+
+        // Step 3. If the player is not holding anything, then harvest the crop.
+        if (heldItem == null) {
+            crop.onHarvest(player);
+            return true;
+        }
+
+        // Step 4. If the held item is excluded from handling, skip it.
```
InfinityRaider: This is the continuation (albeit more elegantly) of my original way of handling this, but I'm still not too fond of it...
Perhaps allow other mods to register excluded items as well?

src/main/java/com/infinityraider/agricraft/farming/CropPlant.java
```diff
@@ -1,129 +0,0 @@
...
-
-import java.util.*;
-import java.util.function.Function;
-
-/**
- * The main class used by TileEntityCrop. Only make one object of this per seed
- * object, and register using the plant registry. ICropPlant is implemented to
- * be able to read data from this class from the API
- */
-public abstract class CropPlant implements IAgriPlant {
```
InfinityRaider: RIP

src/main/java/com/infinityraider/agricraft/renderers/PlantRenderer.java
```diff
@@ -75,7 +75,7 @@ private static void renderCrossPattern(ITessellator tessellator, TextureAtlasSpr
         tessellator.popMatrix();
     }
 
-    private static void renderStemPlant(ITessellator tessellator, TextureAtlasSprite vineIcon, TextureAtlasSprite fruitIcon, int stage, Block vine) {
```
InfinityRaider: I think the Block argument was used to fetch the color (because the vines are always greyscale images, which are then coloured based upon the growth stage)

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
RlonRyan: You get so used to writing public in front of methods that you forget about not needing them in interfaces.

src/main/java/com/infinityraider/agricraft/api/misc/IAgriHarvestProduct.java
RlonRyan: Do you think that this class should exist though? Or should we just let plants have free range on products?

src/main/java/com/infinityraider/agricraft/api/plant/IAgriPlant.java
RlonRyan: I'll have to use a comparator... I removed it since we shouldn't allow sub-classes to be able to change the way the ordering works.

src/main/java/com/infinityraider/agricraft/renderers/PlantRenderer.java
RlonRyan: Yeah, this will have to be reworked. Eventually, the plants should just provide a list of quads for us to render depending on the current state of the plant.

src/main/java/com/infinityraider/agricraft/api/crop/IAgriCrop.java
RlonRyan: Well, there was the thought that a crop might not be a TileEntity, but I suppose that would be near impossible...

src/main/java/com/infinityraider/agricraft/api/crop/IAdditionalCropData.java
RlonRyan: I'm just cutting out everything I can at the moment, making a stable base, and then we can work our way back up from there.

src/main/java/com/infinityraider/agricraft/api/misc/IAgriHarvestProduct.java
RlonRyan: I think you are correct though in that we should change over everything back into weights, since more and more I realize that % chances don't make a whole lot of sense in many cases, although in this case, it is a % chance that the product will be selected as part of any given trial, and the number of trials is determined by the gain stat of the plant.

RlonRyan: And then there is the fact that doubles aren't super efficient in regards to such a limited range.

src/main/java/com/infinityraider/agricraft/api/plant/IAgriPlant.java
InfinityRaider: Good point

PR #967, opened 2017-05-27
Reorganize Imports

RlonRyan:
Since the messy imports were bothering me...


RlonRyan:
I'm just gonna merge this so I can continue further work.

PR #972, opened 2017-05-31
Fix BlockCrop::canBlockStay pos.down() typo

CodesCubesAndCrashes:
Also known as calming down some jumpy crops. An update earlier today in
canBlockStay left off the .down() offset, which meant that the the soil
check would always fail. And since canBlockStay is only called by
neighborChanged, the crops would be stable until an update happened.


RlonRyan:
Good catch, thanks!

Issue #975, opened 2017-05-31
Harvesting Water pad crashes the Server


terorie:
I could reliably reproduce the bug.

My fix: #976 

The server crash occurs when a player tries to break the "Water Pad" (agricraft:water_pad) block.
You get the block by right-clicking farmland with a shovel.

Server context
* Feed The Beast Beyond 1.8.0 Modpack with DraconicEvolution disabled server-side
* Mod state: `UCHIJAAAA	agricraft{2.0.0-0.9.0-a16} [AgriCraft] (agricraft-2.0.0-0.9.0-a16.jar) `
* Commit: [a69faba](https://github.com/AgriCraft/AgriCraft/commit/a69faba7853bc14873b67729a29e5314e337bca6)

This is the crash log.

```
java.lang.IllegalArgumentException: 

   Cannot get property PropertyEnum{
      name=variant, 
      clazz=class net.minecraft.block.BlockDirt$DirtType, 
      values=[dirt, coarse_dirt, podzol]
   } 

   as it does not exist in 

   ExtendedBlockState{
        block=agricraft:water_pad, 
        properties=[powered]
   }

	at net.minecraft.block.state.BlockStateContainer$StateImplementation.func_177229_b(BlockStateContainer.java:196)
	at net.minecraft.block.BlockDirt.func_180651_a(SourceFile:73)
	at net.minecraft.block.Block.getDrops(Block.java:1246)

	at com.infinityraider.agricraft.blocks.BlockWaterPad.getDrops(BlockWaterPad.java:80)

	at net.minecraft.block.Block.func_180653_a(Block.java:544)
	at net.minecraft.block.Block.func_176226_b(Block.java:537)
	at net.minecraft.block.Block.func_180657_a(Block.java:712)
	at net.minecraft.server.management.PlayerInteractionManager.func_180237_b(PlayerInteractionManager.java:336)
	at net.minecraft.server.management.PlayerInteractionManager.func_73075_a(PlayerInteractionManager.java:113)
	at net.minecraft.entity.player.EntityPlayerMP.func_70071_h_(EntityPlayerMP.java:246)
	at net.minecraft.world.World.func_72866_a(World.java:1964)
	at net.minecraft.world.WorldServer.func_72866_a(WorldServer.java:839)
	at net.minecraft.world.World.func_72870_g(World.java:1934)
	at net.minecraft.world.WorldServer.func_184147_l(WorldServer.java:649)
[...]

PR #976, opened 2017-05-31
Fix Invalid drop

terorie:
Not the best solution but at least the server doesn't crash anymore.

Related issue: #975 

I don't know if this drops podzol or dirt, sorry. (Wanted: Drop podzol)
But at least it fixes the server.


RlonRyan:
Good catch! Thanks!

PR #999, opened 2017-06-11
Fix IllegalArgumentException in AgriHarvestProduct

CodesCubesAndCrashes:
The method `Random.nextInt(int bound)` needs bound to be positive,
namely non-zero. Any plant's products that had `min == max` would
therefore cause exceptions. Since the bound is exclusive anyhow, this
commit adds 1 to the bound.

An extra side effect of this bug is that the exception would interrupt
the `onHarvest` event, and so the growth stage would not get reset. You
could get seeds repeatedly until the chance failed and the plant didn't
try to produce any products.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/impl/v1/AgriHarvestProduct.java
+++ b/src/main/java/com/infinityraider/agricraft/impl/v1/AgriHarvestProduct.java
@@ -64,7 +64,7 @@ public final class AgriHarvestProduct implements IAgriHarvestProduct {
 
     @Override
     public int getAmount(Random rand) {
-        return this.getMinAmount() + rand.nextInt(this.maxAmount - this.minAmount);
+        return this.getMinAmount() + rand.nextInt(1 + this.maxAmount - this.minAmount);
     }
 
     @Override


RlonRyan:
That would fix the issue, however it would mess up the bounds. The better fix would to be to add an if statement checking if the min == max and then just returning the min product amount. However, I plan to delete this class, as it did not come out the way I had hoped so I am closing this PR for now. Thanks for the PR though. Your PR's are always very well done.

CodesCubesAndCrashes:
Sounds good, and sincerely appreciate the compliment. :) I want to ask about the bounds though, so that I'm sure I understand. I had looked up the doc for `nextInt`, and it said: 
> [**nextInt**](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt-int-) (int bound)
> Returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (exclusive), drawn from this random number generator's sequence.

So, as I understand it, since the `bound` parameter is exclusive, it needs to be more one higher than your largest desired potential output value. Quick example. Let's say that the min is 3, and the max is 4.
```java
this.getMinAmount() + rand.nextInt(1 + this.maxAmount - this.minAmount)
3 + rand.nextInt(1 + 4 - 3)
3 + rand.nextInt(2)
3 + [0 or 1]
[3 or 4]
```
If you have the moment to point me to where I made the mistake, I would appreciate it very much. Maybe the plant product's max amount is intended to be exclusive too. (But then min should never equal max?) Anyhow, when I tested the patch in game, I thought I was getting the right number of resources when I harvested them. So I'm hoping to learn the right definition, regardless of this class itself. Thank you in advance, and cheers!

PR #1000, opened 2017-06-11
Fix sprinkler particle missing texture and FileNotFoundException

CodesCubesAndCrashes:
This is a two part patch. 

- The first and simpler edit corrects the path used in the ResourceLocation so that the texture is correctly found. This avoids the exception that would occur when the client tried to load up a texture from there. This also fixes the texture on the particles, so that they're not purple-and-black anymore. This however does not fix the incorrect texture on any block breaking particles. 
- The second edit looks bigger, but it is only really adding a call to `bindTexture(TextureMap.LOCATION_BLOCKS_TEXTURE);` once the LiquidSprayFX particle finishes rendering. I even think L26 is unnecessary. The bulk of the edit adds back a few imports that normally would already be present on the `Particle` class. This approach also avoids any changes to AgriCraftFX, to minimize impact. 

### See water and breaking particles in the same scene 

[Particle Texture Troubleshooting Reference](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/modification-development/1437517-1-6-4-custom-particle-textures-overwriting-other)

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/renderers/particles/LiquidSprayFX.java
+++ b/src/main/java/com/infinityraider/agricraft/renderers/particles/LiquidSprayFX.java
@@ -2,6 +2,10 @@ package com.infinityraider.agricraft.renderers.particles;
 
 //heavily inspired by the OpenBlocks sprinkler
 import com.infinityraider.agricraft.utility.BaseIcons;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.renderer.VertexBuffer;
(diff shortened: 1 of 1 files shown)


InfinityRaider:
As always, a nicely made and documented PR, cheers!

CodesCubesAndCrashes:
@InfinityRaider Thank you! :grin:

RlonRyan:
Looks good enough for me!

Issue #989, opened 2017-06-06
Shift right clicking cropsticks to place the cross sticks crashes server


scalda:
Hi, 

I have just updated agricraft and infiniraider

latest version
agricraft-2.0.0-0.10.0-a18.jar
infinitylib-0.10.0.jar

Whenever I place cross sticks via shift right clicking on tilled soil it crashes the server
[crahs report](https://pastebin.com/qz4eHZaK)

regards

Scalda
(FoolCraft Modpack Dev)

RlonRyan:
Yay! Fatal error in fancy new code... And I thought I had gotten better at this...

RlonRyan:
Wait... Wat... How can you get an abstract method error at runtime? Something seriously hinky is going on here... I'm not even sure if that can be debugged...

Uh @tehnut HALP!

CodesCubesAndCrashes:
**Correction:** I'm going to think aloud about something that just occurred to me? In IAgriCrop, getWorld() is declared with default access, not public. That means the method signature is different from TileEntity.getWorld(), which *is* public. I feel like that should still be a compile time issue, but seems worth looking in to. 

#### Original post follows... 

I'm tidying up my research tabs, and since it felt like the discussions on these pages might be relevant, I wanted to share them. 
* [Oracle](http://bugs.java.com/view_bug.do?bug_id=6519118), bug report about calling inherited methods from abstract classes in separate packages. Hard to say if this is similar enough. 
* [Artima](http://www.artima.com/objectsandjava/webuscript/PolymorphismInterfaces1.html), see Static versus Dynamic Binding. I thought this was going to be helpful, since the `liq.gurgle()` example is about unexpected method resolution. But really the bug in agricraft isn't one of the three exceptional cases they list. 
* [StackOverflow](https://stackoverflow.com/questions/32839503/why-does-a-superclass-implement-interface-methods-which-are-supposed-to-be-imple), confirms that a child can fulfill an interface using methods inherited from an abstract parent. So that's indeed supposed to work fine. 
* [Apprize](http://apprize.info/javascript/action/10.html), focuses on default methods in interfaces, but does discuss well the resolution rules.

RlonRyan:
I guess I could rename the method and provide an implementation in the class. The dispatch doesn't make any sense though... And why is this issue just showing up now?

Redhandjane:
[...]

PR #1006, opened 2017-06-15
Rename IAgriCrop methods with AbstractMethodErrors

CodesCubesAndCrashes:
Demonstration album: http://imgur.com/a/MvBJ6

For the IAgriCrop interface, this renames the getPos and getWorld
methods so as to not match the methods inherited from TileEntity.
The only uses so far were in the mutation strategies. And the newly
named methods just call the original named methods via 'this'.

This workaround is for a re-obfuscation bug in Special Source, a
program that ForgeGradle depends on.

https://github.com/MinecraftForge/ForgeGradle/issues/205
https://github.com/md-5/SpecialSource/issues/12

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/api/v1/crop/IAgriCrop.java
+++ b/src/main/java/com/infinityraider/agricraft/api/v1/crop/IAgriCrop.java
@@ -24,17 +24,21 @@ public interface IAgriCrop extends IAgriSeedProvider, IAgriSeedAcceptor, IAgriFe
 
     /**
      * Retrieves the location of the crop instance.
+     * Implementors can just call getPos() if possible. Renamed because of ForgeGradle obfuscation bug.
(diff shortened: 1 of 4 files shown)


CodesCubesAndCrashes:
Fixes #989

RlonRyan:
This looks pretty good... I'm just debating what the methods should be named. It's tricky since the standard method names are taken, so...

I was kinda favoring getCropWorld() and getCropPos() given that the methods are inherited from the IAgriCrop interface. I think that the deobf part of the method name might be confusing to API users...

CodesCubesAndCrashes:
The names weren't easy to pick. My choice was based on starting the same as `getWorld()` and `getPos()`, so that muscle memory and auto-complete would kick in. 

I'll switch them, your names are self explanatory at least. Thanks!

RlonRyan:
I'm just going to make these changes on a commit directly applied to master to avoid all the merge stuff.

"Merged" in a roundabout way.

PR #1007, opened 2017-06-15
Debug blockrange

CodesCubesAndCrashes:
"Hm, why can't this branch automatically merge anymore?" 
"Weird, I don't remember submitting that PR yet..."
"Oh."
Hah. So, this is rather funny. :)

### Quick screenshots: 
testblockrange demo 05

testblockrange demo 07

testblockrange demo 11

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
+++ b/src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
@@ -162,10 +162,10 @@ public class BlockRange implements Iterable<BlockPos> {
             // Post-Increment
             this.x += 1;
             if (this.x > this.range.getMaxX()) {
-                this.x -= 1;
+                this.x = this.range.getMinX();
                 this.y += 1;
                 if (this.y > this.range.getMaxY()) {
-                    this.y -= 1;
+                    this.y = this.range.getMinY();
                     this.z += 1;
                 }
             }
--- a/src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
+++ b/src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
@@ -160,13 +160,13 @@ public class BlockRange implements Iterable<BlockPos> {
             final BlockPos pos = new BlockPos(x, y, z);
 
             // Post-Increment
-            this.x += 1;
-            if (this.x > this.range.getMaxX()) {
-                this.x = this.range.getMinX();
-                this.y += 1;
-                if (this.y > this.range.getMaxY()) {
-                    this.y = this.range.getMinY();
-                    this.z += 1;
+            this.x = this.x + 1;
+            if (this.x > this.range.maxX) {
+                this.x = this.range.minX;
+                this.y = this.y + 1;
+                if (this.y > this.range.maxY) {
+                    this.y = this.range.minY;
+                    this.z = this.z + 1;
                 }
             }
 
--- a/src/main/java/com/infinityraider/agricraft/items/modes/DebugModeTestBlockRange.java
+++ b/src/main/java/com/infinityraider/agricraft/items/modes/DebugModeTestBlockRange.java
@@ -1,0 +1,145 @@
+/*
+ */
+package com.infinityraider.agricraft.items.modes;
+
+// Debug Additions
+import com.infinityraider.agricraft.api.v1.util.BlockRange;
+import net.minecraft.block.BlockColored;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.init.Blocks;
+import net.minecraft.item.EnumDyeColor;
+import net.minecraft.nbt.NBTTagCompound;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+
+import com.infinityraider.infinitylib.utility.debug.DebugMode;
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.EnumHand;
+import net.minecraft.util.math.BlockPos;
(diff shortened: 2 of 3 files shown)


CodesCubesAndCrashes:
I hope you all will forgive me for hitting submit despite the redundancy, hehe. The timing really is funny though, I had finished the docs a couple hours ago, so... jinx?

Anywho, this really was primarily an exercise for myself to write the in-game testing tool. I figured that just reading two `BlockPos`s, looping over the BlockRangeIterator, and placing down some blocks would be very easy. It turned into a lesson in server<->client updates, and the different threads sharing the same objects and values, even if you don't want them to. :grin: 

The bonus would be to fix a small and old bug. And I would be able to use the tool create some screenshots to demonstrate what I had poorly tried to explain with ASCII on the [commit a while back](https://github.com/CodesCubesAndCrashes/AgriCraft/commit/524421230ab09412419b067c5e5f4cc1f1857d07). Wooo. 

#### Well, if anyone has any advice on desyncs and variable storage, I'd love to hear them. 
* For the longest time I would get invisible blocks, since I was trying to only place them on the server-side. For a while I also let the client place their own identical blocks, but I knew that was wrong. Eventually I got it to work by adding `world.notifyBlockUpdate(target, old, wool, 2);` But I don't understand why that's different from `flag:2` inside `setBlockState`. The latter method calls its own notify methods, but it's hard to trace how or if they overlap. 
* The other big problem I had was trying to have the server and client threads maintain their own copies of a variable. Problem of course was that if the variable was on the Debug object, there's only one copy that's shared between both threads. But it turns out that even the `ItemStack` for the Debugger was shared by both threads. Now I *didn't* want to pass information back and forth, but it also seemed impossible to keep the two sides from colliding. Not without having duplicate variables with different names. And those would meanwhile be unnecessary on a pure server. Hm.

InfinityRaider:
I admit, it is a bit tricky if you want to save data between consecutive debug-right clicks, I remember struggling with it too in another mod.

Eventually I just said "fuck it" and only used it for testing client side:
https://github.com/InfinityRaider/SettlerCraft/blob/master/src/main/java/com/InfinityRaider/settlercraft/utility/debug/DebugModePathfinding.java

RlonRyan:
Can you rebase this onto master for me? I like the added debug mode, as the more debug modes the merrier.

RlonRyan:
What I mean to say is can you bring your branch up-to-date, please, so that I might merge it?

CodesCubesAndCrashes:
Sure! Thank you, didn't expect that. :grin:

src/main/java/com/infinityraider/agricraft/items/modes/DebugModeTestBlockRange.java
```diff
@@ -0,0 +1,149 @@
...
+            stack.setTagCompound(tag);
+        } else {
+            tag = stack.getTagCompound();
+        }
+        Optional<BlockPos> start = Optional.empty();
+        assert tag != null;
+        if(tag.hasKey(NBT_START)) {
+            int[] raw = (tag.getIntArray(NBT_START));
+            if (raw.length == 3) {
+                BlockPos p = new BlockPos(raw[0], raw[1], raw[2]);
```
CodesCubesAndCrashes: Oops, now I remember about [`NBTHelper.getCoordsFromNBT()`](https://github.com/AgriCraft/AgriCraft/blob/master/src/main/java/com/infinityraider/agricraft/utility/NBTHelper.java#L23). Just gotta watch out for NPEs. I can shorten this code to use that method instead, if preferred, no worry.

src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
```diff
@@ -160,13 +160,13 @@ public BlockPos next() {
...
-                if (this.y > this.range.getMaxY()) {
-                    this.y -= 1;
-                    this.z += 1;
+            this.x = this.x + 1;
+            if (this.x > this.range.maxX) {
+                this.x = this.range.minX;
+                this.y = this.y + 1;
+                if (this.y > this.range.maxY) {
+                    this.y = this.range.minY;
+                    this.z = this.z + 1;
```
RlonRyan: Why is this registering as changed?

RlonRyan:
Merged into master using command-line.

CodesCubesAndCrashes:
That -is- weird. I'll play with git to see what's up. But it was copied
from AgriCraft/master.

Thank you for the merge!

src/main/java/com/infinityraider/agricraft/api/v1/util/BlockRange.java
InfinityRaider: from "+"= to "+1"

RlonRyan: No I thought he rebased the branch so this should have been matching master.

Kaijiro: "+= 1" could be changed to "++" :)

CodesCubesAndCrashes: I think I can guess why. I ~~rebased the commits and~~ pushed them, before starting typing up the PR. RlonRyan added his fix meanwhile. Since my commits were already public, I didn't rebase them again, but added a final commit to make them match the current master branch. 

What I think github is showing the sum changes of the commits relative to common ancestor? In this case it was ~~master~1~~. (Correction! This branch was relative to 2e227f4. It just didn't conflict with almost anything.)

Tellingly, if you look at the final commit on AgriCraft/master, then you *don't* see this phantom "change".: https://github.com/AgriCraft/AgriCraft/commit/c34d0c3b980ed41c0de7daac817c0435be972de2

**So, a question about procedure:** is it actually okay to rewrite the commits in a PR? To do a rebase or cherry-pick? Will the PR update accordingly, even though the SHA1 don't match anymore?

PR #973, opened 2017-05-31
Fixing up sprinklers

CodesCubesAndCrashes:
A collection of fixes that return basic functionality to sprinklers. Hopefully helps with #862. The commits should be broken up such that a subset of them can still be used even if not all the commits meet the quality and style standards. At the very least, I've tested my patches at every step, and I've stepped through the debugger to confirm both the bugs and the fixes. 

### Summary of changes: 
* The TileEntitySprinkler uses a BlockRange to represent its area of effect. There were a couple problems though: the area was being created above the BlockPosition (not below), and it was created while `pos = BlockPos.ORIGIN`. So all the sprinklers in the world were magically irrigating the same part of bedrock. 
  * I flipped the offset, and I moved the `range` creation to the update method, since the game already calculates a lot of BlockPositions. 
* BlockRangeIterator itself had a logic typo that prevented the full volume from being irrigated. If you look at [this](https://github.com/AgriCraft/AgriCraft/commit/4911415b93db77228606f6ba480aa3b4f5bb9e84#diff-2589ee730ff7f2a571892550a82f190e) commit, you can see the earlier/original full iteration. This was something I particularly checked in the debugger to be sure it really was happening. 
  * I corrected the counters in the loops, and also tidied up the flow. 
* The water channels were looking for sprinklers above, instead of below them. Plus the sprinklers were checking the buffer of the channel, instead of their own. 
  * Flipped the offset for the channels, and replaced the sprinkler's water check. 
* Lastly several of the variables, namely water usage rate, growth rate, and growth chance, were using values not accessible from the config menu. This made them hard to debug. 
  * I exchanged the variables used back for now, so that the sprinklers are easier to test. 

### Quirks and things that still need to be done: 
* The sprinklers irrigate all 245 blocks in the cuboid every time, even if some are buried underground! Oops. So that needs to be replaced with something more realistic and efficient. But since the unrealistic behavior has been around for a while (see this [2b3a5d6] example from 2.5 years ago), I figured fixing that version first would be okay. 
* The rates are using integer math for now, and so the configs aren't as precise as they claim to be. 

### Reference photos: 
* In the first photo, you can see that the farmland is being irrigated in a 7x5x7 cuboid centered below the sprinkler. The tilled earth is dry on the same y-level as the sprinkler block, and also six or more y-levels below. 
agricraft sprinkler patch demo
[...]

Issue #862, opened 2017-03-09
Sprinklers No Longer Work


RlonRyan:
As per the title, Sprinklers no longer work as of the 2.0.0 alphas.

TheDCPlaya:
spinkler doesn't seem to water the farmland
https://clips.twitch.tv/SteamyWanderingLardPlanking

TheChance:
Seems irrigation channels aren't filling at all.

BrewAndCode:
My irrigation channels are filling, I get the whirling particle effect I typically associate with the sprinkler working, however the tilled land is not converting to farmland.
![2017-03-26_22 40 39](https://cloud.githubusercontent.com/assets/6723474/24338692/4d348f60-1275-11e7-8766-6925c176271f.png)

Tsyklop:
Same bug. FTB Beyond 1.10.

Vaygrim:
Can confirm, we are using the latest version of Agricraft and having the exact same issue. Sprinklers might animate and even spew particles periodically, but they don't actually appear to be DOING anything as the channels always step void of water. Bone dry!

Help?

charkko:
Confirmed on 2.0.0a15, Would love to see this fixed on this great mod.

CodesCubesAndCrashes:
Two things that might be causing problems with sprinklers not irrigating correctly. 
* For the area that gets irrigated, the y-offset is flipped, causing the irrigation happening above the sprinkler, not below. [[commit](https://github.com/CodesCubesAndCrashes/AgriCraft/commit/e3f03c803b1b9fe0507d157e1e90190825c6710d)] 
* And I *think* the BlockRange iterator is not fully covering the volume defined by the pair of BlockPos in the two opposite corners. [[commit](https://github.com/CodesCubesAndCrashes/AgriCraft/commit/524421230ab09412419b067c5e5f4cc1f1857d07)]  

I hope I'm not making any mistakes here, and thank you for any feedback in advance.

InfinityRaider:
I dont thinkt he blokrange is a good thing for the sprinkler, because they should stop irrigating once they reach a non "permeable" block

CodesCubesAndCrashes:
Hmm, I was wondering about that too actually. It looks like that has been the effective behavior for a while, so I guessed it was just accepted to work that way. It'd be cool if it was somewhat more realistic. 

[...]

PR #1008, opened 2017-06-15
Improving the updates and irrigation for Sprinklers

CodesCubesAndCrashes:
This new pull request hopefully implements the feedback provided in two previous threads: 
* https://github.com/AgriCraft/AgriCraft/pull/973#issuecomment-306102765 
* https://github.com/AgriCraft/AgriCraft/issues/862#issuecomment-302941453

Consolidated features to be edited in here.

There's also been discussion about even more intelligent looping and tracking, but this version works well already, and can be replaced later.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
+++ b/src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
@@ -4,7 +4,6 @@ import com.agricraft.agricore.core.AgriCore;
 import com.infinityraider.agricraft.api.v1.irrigation.IConnectable;
 import com.infinityraider.agricraft.api.v1.irrigation.IIrrigationComponent;
 import com.infinityraider.agricraft.api.v1.irrigation.IrrigationConnectionType;
-import com.infinityraider.agricraft.api.v1.util.BlockRange;
 import com.infinityraider.agricraft.api.v1.misc.IAgriDisplayable;
 import com.infinityraider.agricraft.blocks.irrigation.BlockWaterChannel;
 import com.infinityraider.agricraft.reference.AgriCraftConfig;
@@ -32,19 +31,26 @@ import java.util.function.Consumer;
 
 public class TileEntitySprinkler extends TileEntityBase implements ITickable, IIrrigationComponent, IAgriDisplayable {
 
-    public static final int BUFFER_CAP = 100;
-    public static final int RADIUS = 3;
-    public static final int HEIGHT = 5;
-
-    private int buffer = 0;
     private int counter = 0;
     private float angle = 0.0F;
-    private boolean active = false;
-
-    private final BlockRange range;
+    private static final int BUFFER_CAP        = 100;
+    private static final int TICKS_PER_SECOND  = 20;
+    private static final int COVERAGE_HEIGHT   = 5; // Configure here. Note: the lowest y-level will be farmland only.
+    private static final int COVERAGE_RADIUS   = 3; // Configure here.
+    private static final int COVERAGE_DIAMETER = 1 + 2 * COVERAGE_RADIUS;
+    private static final int COVERAGE_AREA     = COVERAGE_DIAMETER * COVERAGE_DIAMETER;
+    private boolean active;
+    private int buffer;
+    private int columnCounter;
+    private int waterUsageRemainingMb;
+    private int waterUsageRemainingTicks;
 
     public TileEntitySprinkler() {
-        this.range = new BlockRange(this.getPos().add(-RADIUS, 1, -RADIUS), this.getPos().add(RADIUS, HEIGHT, RADIUS));
+        this.active                   = false;
+        this.buffer                   = 0;
+        this.columnCounter            = 0;
+        this.waterUsageRemainingMb    = Integer.MAX_VALUE;
+        this.waterUsageRemainingTicks = 0;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -70,30 +86,115 @@ public void readTileNBT(NBTTagCompound tag) {
...
+          if (block.isAir(state, this.getWorld(), target)) {
+            continue;
+          }
+          // Option B: Give plants a chance to grow, and then continue onward to irrigate the farmland too.
+          if (((block instanceof IPlantable) || (block instanceof IGrowable))
+              && (this.getWorld().rand.nextInt(100) < AgriCraftConfig.sprinklerGrowthChance)) {
+            block.updateTick(this.getWorld(), target, state, this.getWorld().rand);
+            continue;
+          }
+          // Option C: Dry farmland gets set as moist.
```
MidnightsFX: ++ points for separating watering and growth. Should prevent farmland from potentially getting unlucky and not being watered!

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -70,30 +86,115 @@ public void readTileNBT(NBTTagCompound tag) {
...
+          if (block instanceof BlockFarmland && state.getValue(BlockFarmland.MOISTURE) < 7) {
+            // Note: flags=2 instead of 6, because we're already updating less frequently.
+            this.getWorld().setBlockState(target, state.withProperty(BlockFarmland.MOISTURE, 7), 2);
+            break; // Unnecessary, but explicitly expresses the intent to stop.
+          }
```
MidnightsFX: ++ points for separating watering and growth. Should prevent farmland from potentially getting unlucky and not being watered!

CodesCubesAndCrashes: The current update logic calls `irrigate` on every affected block, so the chance result of any plants doesn't affect the farmland watering currently. But I did see the importance of not changing that.

InfinityRaider:
Haven't looked at it in depth yet, but looks good so far.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -26,22 +27,28 @@
...
+public class TileEntitySprinkler extends TileEntityBase implements ITickable, IIrrigationComponent, IAgriDisplayable {
 
     private int buffer = 0;
     private int counter = 0;
     private float angle = 0.0F;
     private boolean active = false;
-
-    private final BlockRange range;
+    public static final int BUFFER_CAP = 100;
+    private final int TICKS_PER_SECOND = 20;
```
RlonRyan: This should also be static, as there is no need for a new instance of it in every instance of the class.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -56,44 +63,130 @@ public float getAngle() {
     //this saves the data on the tile entity
     @Override
     public void writeTileNBT(NBTTagCompound tag) {
-        if (this.counter > 0) {
+        if (this.counter > 0)
```
RlonRyan: Always put braces after if statements, if you would please. I find it far too easy to add a line to the if statement and cause a different line to fall out of the if statement without noticing.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -56,44 +63,130 @@ public float getAngle() {
...
-            this.counter = 0;
-        }
-
-        if (tag.hasKey(AgriNBT.ACTIVE)) {
-            this.active = tag.getBoolean(AgriNBT.ACTIVE);
-        } else {
-            this.active = false;
-        }
+        this.counter
+            = tag.hasKey(    AgriNBT.LEVEL)
```
RlonRyan: That's definitely a novel way of formatting the compound if statements...

Also, in my experience tag.getInteger() seems to return 0 if no value is found, so these if statements might not be needed. This does outline another reason as to rework the StackHelper class, however.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -56,44 +63,130 @@ public float getAngle() {
...
+        this.waterUsageRemainingTicks
+            = tag.hasKey(    AgriNBT.WATER_USAGE_REMAINING_TICKS)
+            ? tag.getInteger(AgriNBT.WATER_USAGE_REMAINING_TICKS)
+            : 0;
     }
 
     //checks if the sprinkler is CONNECTED to an irrigation channel
     public boolean isConnected() {
-        return this.worldObj != null && this.worldObj.getBlockState(getPos().add(0, 1, 0)).getBlock() instanceof BlockWaterChannel;
+        return this.worldObj != null && this.worldObj.getBlockState(getPos().up()).getBlock() instanceof BlockWaterChannel;
```
RlonRyan: Those direction methods sure were a great addition to the BlockPos class, that's for sure!

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -56,44 +63,130 @@ public float getAngle() {
...
+            // Step 1: Check if we need to refresh the water usage variables
+            if (waterUsageRemainingTicks <= 0) {
+                this.waterUsageRemainingMillibuckets = AgriCraftConfig.sprinklerRatePerSecond;
+                this.waterUsageRemainingTicks = TICKS_PER_SECOND;
+            }
+
+            // Step 2: Update the per-tick rate.
+            // Note: This math executes regardless of actual consumption occuring, so
+            //       that the usage rate refresh still happens every twenty ticks.
+            final int waterUsageThisTick = this.waterUsageRemainingMillibuckets / this.waterUsageRemainingTicks;
```
RlonRyan: Personally, I find it best to avoid division in code wherever possible since it can wind up being inaccurate, and you have to worry about casting too.

RlonRyan: I'm not sure if this can be done without division though, so...

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -56,44 +63,130 @@ public float getAngle() {
...
+            if (this.active) {
+                this.buffer -= waterUsageThisTick;
+                doSprinkle();
+            }
         } else if (this.active) {
             this.renderLiquidSpray();
         }
     }
 
+    public void doSprinkle() {
```
RlonRyan: I would reccomend taking a peek at the WorldHelper class if you continue work on AgriCraft. I've found it to be most useful, and it decreases code duplication homogenizing the codebase as a result.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
```diff
@@ -234,4 +292,10 @@ private void spawnLiquidSpray(double xOffset, double zOffset, Vec3d vector) {
         Minecraft.getMinecraft().effectRenderer.addEffect(liquidSpray);
     }
 
+    @Override
+    @SideOnly(Side.CLIENT)
+    public void addDisplayInfo(Consumer<String> information) {
+        // Borrowed from TileEntityChannel.
```
RlonRyan: Is there some importance to this comment?

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Oops! Yup. I overlooked that during review.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
TehNut: Yeah, all NBT getters have nonnull defaults.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Will do. I had been wondering about it. And I'll fix all the instances of that here.

CodesCubesAndCrashes: A related question occurs to me: can these `if` tests cause problems if this is called on an pre-existing tag with old data on it, that might not get overwritten with the default values. The `writeToNBT()` method gets called a lot, so although all the ones I saw were either called on brand new tags, or delegates to their super class, it's hard to be sure. Maybe they should all have else clauses that remove the corresponding tag if it's there? How important is it to avoid writing this data when it's equal to the default?

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Yup. I don't know where the idea came from. I like a lot how it makes it easier to visually scan the keys and alternative values. I don't like how the line with the `?` isn't the test expression, and the line that starts with `=` could be misleading. :confused: 

Looking at NBTCompoundTag.java, the javadocs for the getters all specify default values, yup. Are the docs from Mojang or Forge? Are they stable? Is relying on the defaults good practice? Sorry for the tedious questions. 

I thought it might be possible to add a generic helper method to NBTHelper.java
```java
public static <T> T getKeyOrElse(String key, T backup) { ...
```
But you can't use `instanceof` on backup to identify the real getter. Method overloading is always possible, but don't know if it's worth it.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Quick check: is inaccuracy a concern if this is doing integer/truncated division intentionally? Should there be a comment noting that? Is there some way to annotate it for the compiler?

Say that the rate is 38 mB/sec. Subtract the third column from the first to get the next row. 

mB remaining | ticks remaining | mb/t =
--------------- | ---------------- | --------
38 | 20 | 1
37 | 19 | 1
36 | 18 | 2
34 | 17 | 2
... | ... | ...
4 | 2 | 2
2 | 1 | 2
38 | 20 | 1
... | ... | ...

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Finding and reading the tools or utilities was/is fun. :) Still re-read them periodically to see new ways they can help.

Mm, that having been said, can I get a hint about what you see could be replaced and simplified?

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Attribution. Not passing code off as my own. Habit? Also, when I was making that patch, I documented the source so that if there were problems using the same code in the new place, it could be easier to figure out what the mistake was in applying it. But, in hindsight, a note like this is better in a commit message. So the answer to your question is: "nope!"

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: Self-documenting FTW.

CodesCubesAndCrashes:
Revisions have been made as suggested or requested, except where I asked questions. 

Random question: I've found `git rebase --interactive HEAD~4` (or whatever number needed) to be handy and nice for tidying things up. But once I've pushed to the public facing github, I shouldn't rewrite that history, I've understood. Is that true for a PR too? Should I eventually do a rebase to tidy up or squash some commits? Or will it not matter, because if a PR gets accepted, then the merge commit will also flatten everything.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
RlonRyan: Well, the inaccuracy is relative. If you prove that the formula works like you have done above, then there is nothing to worry about. It really just makes the computations harder to prove.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
RlonRyan: I've debated the same thing. The issue is that if you use the generic method then the primitives wind up getting wrapped. I think you just have to settle with hard coding method variants for each primitive type with the addition of the String type.

mod.properties
```diff
@@ -11,7 +11,7 @@ package = com.infinityraider.agricraft
 # Version
 version_major = 2
 version_minor = 0
-version_patch = 0-0.10.0-a18
+version_patch = 0-0.10.0-a19-in-development
```
RlonRyan: Perhaps don't commit this change? It might conflict if I wind up doing a release before this PR gets merged...

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: I went ahead and switched all but one of them to using the default values of the getter methods. I added a note to document those expectations, in case the behavior ever changes upstream.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: It's been removed as part of the squash.

mod.properties
CodesCubesAndCrashes: This change has been removed now.

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: If you check line 103, I found an opportunity to simplify the check using `WorldHelper`. Thanks!

src/main/java/com/infinityraider/agricraft/tiles/irrigation/TileEntitySprinkler.java
CodesCubesAndCrashes: ### Well, we can ignore most of the math and focus on just these facts: 

* The `remaining ticks` starts at 20. 
* During each update, the `remaining water` amount shrinks by `remaining water / remaining ticks`, but let's just call it `"some amount"`.
* Also, during each update the amount of `remaining ticks` shrinks by 1. 
* On the twentieth update, the `remaining ticks` will be equal to 1, and so that `"some amount"` is just `remaining water / 1`. 
* Therefore, when the `remaining water` amount shrinks by `"some amount"`, then really it shrinks by itself. 
* And that means that on the twenty-first update, both `remaining water` and `remaining ticks` will be equal to zero. Which means the previous second is done, and we are now doing the next second!

Bottom line is: even if there is a bug along the way in java's integer division, the total amount of water consumed every second will be the desired amount. 

Plus: this code doesn't use any floating points, so both the performance hit and the headaches from dealing with casting are both already avoided. :)

### On the other hand, here's the math: 
#### I've been trying to think of what would be a satisfying proof. Hopefully this succeeds a little bit.

Firstly, in the case when `sprinklerRatePerSecond % 20 == 0`, then trivially 
* `waterUsageThisTick = sprinklerRatePerSecond / 20`. 

If instead `sRPS % 20 != 0`, then let us start by defining that:
* `waterUsageRemainingMB_t0 = sRPS`, 
* `waterUsageRemainingTicks_t0 = 20`, 
* `r_tN = wURMB_tN % wURT_tN`, and 
* `q_tN =  (wURMB_tN - r_tN) / wURT_tN`. Phew! 

Now, in normal math, `sRPS / 20.0` has a non-integer answer. But in this line of code we are using integer division (*idiv*, see below). That means: 
* `waterUsageThisTick_tN = q_tN`,
* `wURMB_tN+1 = wURMB_tN - wUTT_tN`, and
* `wURT_tN+1 = (wURT_tN) + 1`. Heh. 

Now, since *idiv* rounds towards zero, then the quotient for integer division will *always* be less than or equal to the quotient for real/normal division. And as long as we are shrinking the remaining amount of mB by a value less than the real quotient, then logically the real quotient can only increase on successive iterations, until there is no integer remainder. After that, the integer quotient will be equal to the real quotient, and so the real quotient will stop growing. 

> [The JVM Specification, Java SE 8 Edition, Page 453: **idiv**](
https://books.google.com/books?id=0o-AAwAAQBAJ&pg=PA453) 
> An `int` division rounds towards zero; that is, the quotient produced for `int` values in *n/d* is an `int` value *q* whose magnitude is as large as possible while satisfying |*d* * *q*| <= |*n*|. 

Summarizing: if you shrink at a rate less than your actual remaining average rate, then your remaining average rate will have to grow to make up for it. Once the remaining average rate is a whole/integer number, then you can match integer consumption rate to the remaining average rate, and so the remaining average rate won't change any further.

darthvader45:
Okay, seems that both this and another pull request passed checking.  Can't wait to see the next iteration of this mod.  Keep up the awesome work so far.

PR #1017, opened 2017-06-28
Fixed Mutations

bibo38:
Hi guys,

just stumbled on this bug, while trying to get my Sugarcane seed.

To get the stats of the mutation, the StatCalculatorRegistry is asked
with an IAgriMutation object. But the default StatCalculatorRegistries
only handle the IAgriPlant interface. Therefore the code fails to
calculate the stats and no mutation will occur.


RlonRyan:
Good catch! Thanks for the PR!

PR #1021, opened 2017-06-30
Fix plant products to be in min,max range

CodesCubesAndCrashes:
This changes which toStack method gets called by getHarvestProducts.
When the Random parameter was missing, AgriStack.toStack(Class) was
being called on the parent, and not AgriProduct.toStack(Class, Random).
The harvest amount randomization from min to max occurs in the latter.


RlonRyan:
Well that was dumb of me...

CodesCubesAndCrashes:
For what it's worth, I made the exact same mistake myself! On the discord server I had claimed to dilnu that this was fixed, and then described all the steps that should be happening. Turns out that I was only assuming that it was doing what I thought. Mostly because I saw the method names matched, it compiled, and the IDE didn't point out anything weird. Plus, I have a memory of testing a different fix for this bug long ago, and that mislead me to think I had tested a18 too. Oops.... 😒

PR #1022, opened 2017-06-30
Prevent vanilla bonemeal effect on empty crops

CodesCubesAndCrashes:
This pull request consists of two different fixes, either of which resolves the issue, but both of which I think are worth keeping anyhow. Fixing canGrow will prevent other mods and methods from getting misleading results and wasting resources by directly calling applyBonemeal. And fixing onBlockActivated will prevent similar bugs from arising when fertilizers other than bonemeal were integrated into AgriCraft.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
+++ b/src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
@@ -132,8 +132,10 @@ public class BlockCrop extends BlockTileCustomRenderedBase<TileEntityCrop> imple
 
         // Step 5. If the held item is a type of fertilizer, apply it.
         if (AgriApi.getFertilizerRegistry().hasAdapter(heldItem)) {
-            Optional<IAgriFertilizer> fert = AgriApi.getFertilizerRegistry().valueOf(heldItem);
-            return fert.isPresent() && fert.get().applyFertilizer(player, world, pos, crop, heldItem, crop.getRandom());
+            AgriApi.getFertilizerRegistry()
+                    .valueOf(heldItem)
+                    .ifPresent(f -> f.applyFertilizer(player, world, pos, crop, heldItem, crop.getRandom()));
+            return true; // Even if it didn't grow or cross, we've still handled this block activation, so return true.
         }
 
         // Step 6. If the held item is crops, attempt to make cross-crops.
--- a/src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
+++ b/src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
@@ -212,7 +212,7 @@ public class BlockCrop extends BlockTileCustomRenderedBase<TileEntityCrop> imple
      */
     @Override
     public boolean canGrow(World world, BlockPos pos, IBlockState state, boolean isClient) {
-        return this.getCrop(world, pos).map(crop -> !crop.isMature()).orElse(false);
+        return this.getCrop(world, pos).map(crop -> crop.hasSeed() && !crop.isMature()).orElse(false);
     }
 
     /**


InfinityRaider:
I think I decided a long time ago that canGrow returns true to allow for weeds to spawn on empty crops.

CodesCubesAndCrashes:
@InfinityRaider I'm not actually sure anymore how `canGrow` should change, but I'd love any feedback on my observations. 
* Weeds are now normal plants, and spawn via the `spawn()` method, called by `TileEntityCrop.onGrowthTick()`. They don't need special code anymore. 
* I looked at all the implementations of `IGrowable.canGrow()`, and most of them focus on maturity, and the rest spawn new plants nearby. 
* The `IGrowable` interface is only used by ItemDye in vanilla, and `IGrowable.grow()` is called by the crop tile entity. But if you [search for imports](https://github.com/search?q=%22import+net.minecraft.block.IGrowable%3B%22&type=Code&utf8=%E2%9C%93) of it on Github, you can see it's used commonly enough. 
* Right now, using the agricraft fertilizer calls `BonemealWrapper.applyFertilizer`, `TileEntityCrop.acceptsFertilizer`, `TEC.onApplyFertilizer`, [`BlockCrop.grow`](https://github.com/AgriCraft/AgriCraft/blob/master/src/main/java/com/infinityraider/agricraft/tiles/TileEntityCrop.java#L402), [`TEC.onGrowthTick`](https://github.com/AgriCraft/AgriCraft/blob/master/src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java#L235), and finally `TEC.applyGrowthTick`. It seems like the step through BlockCrop is unnecessary, and possibly outdated since it still uses `Constants.MATURE`. 

I can see `IGrowable` being treated one of two ways: 
* Make all three methods return false and do nothing. This could simplify potential bugs. Right now, even if you want to block an item using the `IAgriFertilizer` logic, if it calls `grow` then it will bypass AgriCraft's control. But the result will *look* the same because they both use `onGrowthTick`, so you wouldn't realize that this is where the bug is. Also, then there's no ambiguous compatibility: until someone adds support explicitly, you won't have bugs like 1018 where it looks like it's working but in reality it isn't doing anything. 
* Have `canUseBonemeal` return `true` instead. 
  * You instantly should have support for other mods' fertilizers, if they work on vanilla blocks.
  * If you want to override that behavior, then you add a compatibility adapter to the registry. The block's right click handler gets control before the item's handler, so 

#### Good reference I found much later: http://www.minecraftforge.net/forum/topic/22365-understanding-igrowable/ 
Basically, canGrow is used to decide if a plant is fully grown yet, and grow makes the actual change happen. I think canUseBonemeal is misleading, and badly named: in vanilla, if bonemeal has a chance to fail (on saplings, for example), then cUB is where the randomness happens. If cUB returns false, then bonemeal is still consumed, but no benefit happens.

InfinityRaider:
I'm following your reasoning, by itself that seems the best option indeed.
However in 1.7.10 I remember having to mess around with those three IGrowable methods because they were used by other mods in automated harvesting (to check if a crop is mature etc...), so returning false from all three might be a good solution now, but eventually when looking at compatibility, it might give some issues.

renadi:
It sounds like if they were using a check for something other than what
they actually wanted to know it was probably the wrong way.  Doesn't forge
have some kind of harvestable or fully grown tag that would make more sense
to use?

CodesCubesAndCrashes:
@InfinityRaider I think the best, and reasonably difficult, solution is to make the vanilla methods respect a config setting. If a player wants to be hardcore, they can block any (potentially easier) third party mods from using the vanilla `IGrowable` methods to speed up growth. And otherwise (by default?) it'll accept those method calls. Personally, I think that until AgriCraft has a richer fertilizer mechanic, blocking outside fertilizer/bonemeal effects is more confusing than interesting. I'll put together a PR with an implementation suggestion of this then. Thanks!

src/main/java/com/infinityraider/agricraft/blocks/BlockCrop.java
```diff
@@ -212,7 +214,7 @@ public void dropBlockAsItemWithChance(World world, BlockPos pos, IBlockState sta
      */
     @Override
     public boolean canGrow(World world, BlockPos pos, IBlockState state, boolean isClient) {
-        return this.getCrop(world, pos).map(crop -> !crop.isMature()).orElse(false);
+        return this.getCrop(world, pos).map(crop -> crop.hasSeed() && !crop.isMature()).orElse(false);
```
RlonRyan: I think that a crop cannot be considered mature if it doesn't have a seed, right?

CodesCubesAndCrashes: Correct, but that's also the problem: 
1. If a crop is not mature, then it can grow. 
2. If a crop has no seed, then it is not mature. 
3. Therefore, ***if a crop has no seed, then it can grow***.
4. And... if `IGrowable::canGrow` returns true for a target, then `ItemDye::onItemUse` will consume a bonemeal from the stack, even if `canUseBonemeal` for that target returns false, and no growth tick happens. (For that detail, see my walls of text elsewhere. :P)

----

Woo, haha. At first, your question made me doubt everything I had written and tested. 😀 Because you're right that `isMature` already depends on `hasSeed` being true. But the trick is that there are three states to deal with, not two: no plant, immature plant, and mature plant. And so this edit that adds `hasSeed` here is not redundant. `!isMature` is insufficient. 

Also, according to my memory, I had tested these two commits separately, and each one had fixed the bug independently. 

So, when I looked up `isMature`, and that the answer to your question is yes, I got really confused, and doubted my memory and logic. At least until I walked through it and remembered what the real problem was. 

I hope that this makes sense! Let me know.

CodesCubesAndCrashes:
@RlonRyan I don't think I've directly and clearly asked this elsewhere, so I'm going to now. 

I would like to add on to or replace this pull request with one where the `IGrowable` methods check a new configuration setting, such as: 

> "Can AgriCraft crops grow from mods that use the vanilla growth methods instead of having dedicated compatibility?" (True/False)

The result would be that AgriCraft could natively and for free support the mods that use the typical methods. But also that could be disabled for the hardcore packs and players. 

So, is this something that *you* would like?

RlonRyan:
The other thing is that I had it so that bonemeal could spawn in weeds, to encourage against the indiscriminate use of fertilizer on crops.

CodesCubesAndCrashes:
Mm, I could make sure that that risk still exists. That's a good balance.

This bug however is separate from that. The bonemeal is consumed by the vanilla IGrowable methods, because canGrow returns true, but no growth tick happens because canUseBonemeal returns false. The names are not intuitive, I've learned.

RlonRyan:
Yeah the whole bonemeal thing is a mess. It'd probably be easier if we just came up with our own fertilizer items and ignore others, but then everyone would complain about compatibility. The issue is that the vanilla bonemeal code is not really conducive to extensions.

CodesCubesAndCrashes:
In my testing, it was possible to have the IGrowable interface implemented, letting outside mods trigger growth, while the agricraft IFertilizer still took precedence when valid. That let me do things like block bonemeal in the wrapper and it still wouldn't trigger the vanilla interface.

CodesCubesAndCrashes:
### Retroactive update
Adding the IGrowable interface to `BlockCrop` went fine. I got sidetracked creating a debug item that could cycle between the four combinations of supporting/not-supporting either IGrowable or IAgriFertilizer. That way it would be easier to see what effect enabling or disabling the new config option had. That item turned out to be more complicated than the patch itself, and since the enhancement works already by itself, I'm remembering to put it up for scrutiny first. 

**[The branch is here](https://github.com/CodesCubesAndCrashes/AgriCraft/tree/dev-igrowable)**, if anyone wants to take a look at the logic. I'll rebase it against a20 before making a fresh pull request.

PR #1023, opened 2017-07-01
Feature(s): Already analyzed seeds update the journal. And analysis progress doesn't get reset by removing the journal.

CodesCubesAndCrashes:
When a specimen or journal is inserted into the analyzer, such that an
analyzed seed is present at the same time as a journal that lacks that
plant, then the plant will be added to the journal. This makes the
behavior more intuitive, and less roundabout. Credit for pointing out
this possibility goes to @Dilnu.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/tiles/analyzer/TileEntitySeedAnalyzer.java
+++ b/src/main/java/com/infinityraider/agricraft/tiles/analyzer/TileEntitySeedAnalyzer.java
@@ -288,6 +288,7 @@ public class TileEntitySeedAnalyzer extends TileEntityRotatableBase implements I
                     } else {
                         output = this.specimen.copy();
                         this.specimen = null;
+                        this.progress = 0;
                         this.markForUpdate();
                     }
                 }
@@ -300,7 +301,6 @@ public class TileEntitySeedAnalyzer extends TileEntityRotatableBase implements I
                 }
                 break;
         }
-        this.progress = 0;
         return output;
     }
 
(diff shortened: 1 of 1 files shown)


CodesCubesAndCrashes:
Here's this for now. :)

https://github.com/AgriCraft/AgriCraft/pull/1023

The minimal version of this request can be done with two lines and a bracket, like @InfinityRaider said elsewhere. This implementation uses a more robust approach instead. 

* The order of inserting items doesn't matter. For one example, you can forget to put your journal in the analyzer until after you've scanned a seed. And you can save analyzed seeds and then add them to a journal later. 
* Variables are saved as needed, so that a few chains of method calls aren't getting repeated. 
* The easy or most common conditions are tested first. 
* But, fix, to-do: non analyzed seeds will get added to the journal too. That's a big typo from simplifying things. I'll fix and push to this PR again.

Overall, I don't have an opinion yet about how this affects gameplay, and if it's a good choice in the big picture. But having code to actually play with seems worthwhile. Let me know!

InfinityRaider:
Solid implementation

CodesCubesAndCrashes:
While testing, I noticed that removing the journal from the analyzer, while it was in the process of analysis, would reset the progress to zero. Since the analysis doesn't depend on the journal, I decided it would be more intuitive to reset the progress only if the specimen/seed itself was removed. That's the second commit, even though it's outside the current scope of this PR.

RlonRyan:
Well if @InfinityRaider likes it then I see no reason as why not to merge it!

Issue #1055, opened 2017-08-01
Crop spreading / stat increase wont work ?


Darianna77:
Hello....since weeks ill try to increase the stats from my plants but it isnt possible for me :/
Is there something change to 1.7.10 ?

In 1,7.10 i have made a row with 9 or 10 cropsticks....in the first i plant a seed....and the other all double cropsticks...
have tried it....sometime the plant spreaded 1 more time....but then it stops....have wait 2 or 3 minecraft weaks...nothing
when i plant one crop...then an double cropstick....then another crop. Sometime it spreads in the middle thing...but i get never any stat increase...
When i made an single cropsticks...the crops spread...but wont growth..

So i have tried to change the spreadrate for wheat on 0.5 ...than it spreads.....but the stat wont increase...
have change the config for single spread stat increase to ture also....but all what i get after 10 spreads is 1-1-1

so please help me...whats wrong...what are the difference to 1.7.10 ?
after playing nearly one rl month i really will have better crops :/

MBelhardi:
I can confirm, that there must be something wrong with the spreading values.

Unfortunately, I must say, that spreading the seeds takes much too long. If my thought is correct, and the spreading will be calculated from the general spreading chance and the spreading chance of the individual plant, then that is no wonder. The default value, of the config file, for the spreading chance is **0.15**, and the spreading chance of most plants is **0.1**, which results in a total-spreading chance of **0.15 x 0.1 = 0.015**. At least I think, that this is how the spreading chance will be calculated, otherwise I can't actually explain, why the spreading needs so long.

config-value:
```
    # The base chance for a crossover to occur during any given tick. Setting this 95% is of questionable morality. [range: 0.05 ~ 0.95, default: 0.15]
    S:"Crossover Chance"=0.15

```
plant-value: (for example wheat)
`  "spread_chance": 0.1,
`  

I had to change the config-value to **0.55** and the plant-value to **0.3** (in **total 0.165**) to get a reasonable spreading time. (while I'm using the agricraft irrigation system) I don't want to know, how long it will take, without the irrigation system.
With these changes it took me almost 6 hours to get 1 10/10/10 Seed. (weed was disabled)

Also, sometimes plants will spread to another crop, even though it's not a double crop

cap1o:
Hello !

I have got the same problem with the 1.10.2 Agricraft 2.19a on Beyond modpack.
Sprinklers seem not working and i wait 3H IRL to get nothing more on plant stat.
[...]

Issue #1036, opened 2017-07-09
a19: Potatoes are unable to pass 5/5/5


DinDillian:
Spent the past day trying to pass 5/5/5, and have been unable to so far. The configs are set to 10/10/10.

DinDillian:
Created a modpack with just JEI, Agricraft a19 and Mystical Agriculture and couldn't pass 3/3/3.

CodesCubesAndCrashes:
@DinDillian  
Hi! I've uploaded a preview of the a20 release, which contains the fix for this stat increase problem. If you get a chance to try it out, let me know. Any feedback will be great. :)

https://github.com/AgriCraft/AgriCraft/releases/tag/v2.0.0a20

DinDillian:
@CodesCubesAndCrashes Tried using it and I have successfully managed to get some 10/10/10 potatoes

Issue #1028, opened 2017-07-02
a19: Seeds degrade on single spreading.


Dilnu:
It looks like the mutation stat divisor change might be being applied to sam-crop mutations which is confusing.

CodesCubesAndCrashes:
@Dilnu 
Hi! I've uploaded a preview of the a20 release, which contains the fix for this stat increase problem. If you get a chance to try it out, let me know. Any feedback will be great. :)

https://github.com/AgriCraft/AgriCraft/releases/tag/v2.0.0a20

PR #1062, opened 2017-08-23
Fix calculateStatMeanDivisor invalidParents typo

### LINKED ISSUE #977: Mutation Spread Bug/Mystical Agriculture Bug
Hey I am having a couple issues with agricraft right now. I currently have the latest version installed 2.0.0-0.10.0-a17.
 
Current Bugs
- Wheat crops do not mutate to make better wheat crops when using cropsticks, this could be with other crops as well but I haven't tried any yet.
- ~Mystical Agriculture crops keep spawning inside of my cropsticks at random times~ [known issue]
 
I would really like some help to get this mod working as I enjoy it a lot :) if you could get back to me as soon as possible it would be appreciated

CodesCubesAndCrashes:
The typo in the otherCropsAffectStatsNegatively code meant that
instead of accounting for any invalid parents, the function would
always return double the number of valid parents. This made it
difficult for stats to increase in that case.

-----
Fixes #977 #1028 #1036 and part of #1055.

I have successfully bred plants (wheat, carrots, beets, pumpkins, etc)
to 10/10/10, see screenshot for an example. Importantly, this was with 
the config for `otherCropsAffectStatsNegatively` enabled, since that 
was part of the bug. For these experiments I had `fertilizerMutation`
set to true, spread strategy at 50%, and the stat calculator set to 
normal. It also took several stacks of bonemeal, but overall less than 
15 minutes to get a plant up to max, starting with 1/1/1 seeds.

agricraft spread stats

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/farming/mutation/statcalculator/StatCalculatorBase.java
+++ b/src/main/java/com/infinityraider/agricraft/farming/mutation/statcalculator/StatCalculatorBase.java
@@ -160,7 +160,7 @@ public abstract class StatCalculatorBase implements IAgriStatCalculator, IAgriAd
 
         // Account for invalid parents, if enabled in config.
         if (AgriCraftConfig.otherCropsAffectStatsNegatively) {
-            meanDivisor = meanDivisor + validParents;
+            meanDivisor = meanDivisor + invalidParents;
         }
 
         // Verify that the mean divisor is at least one.


InfinityRaider:
Ok whaw, good catch

CodesCubesAndCrashes:
@InfinityRaider I don't think so. That block of code was introduced in [this commit](https://github.com/AgriCraft/AgriCraft/commit/dd40bd9d5b7f0579058777d8f5ef68258b960ba1) back on May 25th of this year.

CodesCubesAndCrashes:
Thank you for that quick review and response!

PR #1063, opened 2017-08-23
Block trowels from working on cross-crops

CodesCubesAndCrashes:
This edit of `ItemTrowel::onItemUse` adds a check for cross-crop status on
the crop. This prevents trowels from inserting seeds into cross-crops.

----
While testing, I noticed that I could use a full trowel on a cross-crop, and
this would set that crop to have that seed, but leave the crop still as a 
cross-crop. It rendered both aspects successfully, but the crop itself was
now in an invalid state. Since it was a cross, the plant would never get to
grow. And a cross-over would overwrite the seed already inside the crop.
Because of the same bug, the trowel could actually remove the seed from
the cross-crop-with-seed.

#### Edit: 
I noticed too late that an unnecessary import got included in the patch.
It's been removed, the commit's been amended, and the pull request is updated.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/items/ItemTrowel.java
+++ b/src/main/java/com/infinityraider/agricraft/items/ItemTrowel.java
@@ -56,7 +56,10 @@ public class ItemTrowel extends ItemBase implements IAgriTrowelItem, IItemWithMo
         if (te instanceof IAgriCrop) {
             IAgriCrop crop = (IAgriCrop) te;
             Optional<AgriSeed> seed = AgriApi.getSeedRegistry().valueOf(stack);
-            if (crop.hasSeed() && !seed.isPresent()) {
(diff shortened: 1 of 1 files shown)


CodesCubesAndCrashes:
## From testing:
#### The result is self-explanatory enough. Picked up a plant with a trowel, and put it into a cross crop.
agricraft itemtrowel and cross crops 1
#### What's extra fun is combining this with #1064, to get a mature plant inside the cross crop too.
agricraft itemtrowel and cross crops 2
#### Right-click to harvest causes individual crop sticks to spawn, while leaving the crop unchanged.
agricraft itemtrowel and cross crops 3
#### I should probably fix this last thing as a separate bug. Just one line to change in `onHarvest`. It's also unrelated to the mature plant aspect.

InfinityRaider:
Seems good

PR #1064, opened 2017-08-23
Reset growthstage to zero when plants change

CodesCubesAndCrashes:
#### Original commit is here: 840c5f78

-----
#### Commit message:
With this edit:
- TileEntityCrop::setGrowthStage can now reset growth even when there's
  no plant. It also makes sure that growthStage always gets set to zero
  in that case, and logs a warning otherwise.
  It now returns a boolean to indicate if there was a change and update.
  Also it uses MathHelper::inRange instead of doing it manually.
- TileEntityCrop::setSeed now checks first if the new seed is equal to
  old, using the AgriSeed::equals method if possible, and returns early.
  It now calls setGrowthStage(0), and wil only call markForUpdate if
  that method didn't already.
- TileEntityCrop::onRaked doesn't have to reset the growth as a result.
- The signature for IAgriCrop::setGrowthStage was changed to match. The
  existing uses are unaffected, since the return value just gets ignored.
- Documentation was added or updated in IAgriCrop, IAgriSeedAcceptor,
  and TileEntityCrop for all these methods, to record the effects,
  expectations, and intentions of the code.

Overall:
This prevents items (such as the trowel) from removing a grown plant
while leaving the growthStage unchanged. A different plant or seed then
could get planted and it would have that non-zero growthStage.
Similarly, it prevents accidents where code might remove the seed before
resetting the growth stage, instead of after. Setting the growth stage
would silently fail in that scenario.
Finally, it reduces the number of calls to markForUpdate by removing
some existing repetition. (E.g. onRaked.)

----
#### Commentary:
While moving plants around using a trowel, I noticed that extracting a
mature plant from a crop and then inserting it to the same crop would
return it to its mature state, but inserting it into a fresh crop produced
a plant with zero growth. Further testing showed that the growth level
could be transferred between plants this way, allowing for all sorts of
mischief.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/api/v1/seed/IAgriSeedAcceptor.java
+++ b/src/main/java/com/infinityraider/agricraft/api/v1/seed/IAgriSeedAcceptor.java
@@ -20,9 +20,10 @@ public interface IAgriSeedAcceptor {
 
     /**
      * Sets the seed associated with this instance.
+     * Use null to remove the seed from this instance.
      *
      * @param seed the seed to associate with this instance.
-     * @return the seed priorly associated with this instance.
+     * @return true if this changed the value, or false if they were equal already.
      */
     boolean setSeed(@Nullable AgriSeed seed);
 
--- a/src/main/java/com/infinityraider/agricraft/tiles/TileEntityCrop.java
+++ b/src/main/java/com/infinityraider/agricraft/tiles/TileEntityCrop.java
@@ -1,6 +1,7 @@
 package com.infinityraider.agricraft.tiles;
 
 import com.agricraft.agricore.core.AgriCore;
+import com.agricraft.agricore.util.MathHelper;
 import com.infinityraider.agricraft.api.v1.AgriApi;
 import com.infinityraider.agricraft.api.v1.crop.IAgriCrop;
 import com.infinityraider.agricraft.api.v1.fertilizer.IAgriFertilizer;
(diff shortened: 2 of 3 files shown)


src/main/java/com/infinityraider/agricraft/tiles/TileEntityCrop.java
```diff
@@ -193,6 +193,7 @@ public boolean setSeed(AgriSeed seed) {
         final AgriSeed oldSeed = this.seed;
         this.seed = seed;
         this.markForUpdate();
```
InfinityRaider: setGrowthStage() will also call markForUpdate, don't call that twice, it will create overhead

InfinityRaider: ...unless the growth stage already was 0

CodesCubesAndCrashes: I saw that, and went with the more cautious approach. Relying on `setGrowthStage` to call `markForUpdate` seemed like it was asking for a bug later if it got changed.

I could add to the documentation of both to note these behaviors though, that would hopefully be enough.

I was wondering about a related thing: `setSeed` always marks for update, even if the seed isn't different. I don't see anywhere that relies on that quirk, but it might be an intention or assumption anyhow. It would be easy to instead only reset the growth and update the seed if it was different.

CodesCubesAndCrashes:
## Screenshots I took during testing. 
#### First we use a trowel on a crop with a mature plant on it, and get an "empty" crop. 
agricraft itemtrowel and growthstage during
#### Then we plant a seed by regularly right clicking on the crop with the seed. And the resulting plant is mature already. 
agricraft itemtrowel and growthstage after

src/main/java/com/infinityraider/agricraft/api/v1/seed/IAgriSeedAcceptor.java
```diff
@@ -20,9 +20,10 @@
 
     /**
      * Sets the seed associated with this instance.
+     * Use null to remove the seed from this instance.
      *
      * @param seed the seed to associate with this instance.
-     * @return the seed priorly associated with this instance.
+     * @return true if this changed the value, or false if they were equal already.
```
CodesCubesAndCrashes: This file change is purely updating the doc to match the existing behavior. It is unaffected by the rest of this commit actually.

PR #1065, opened 2017-08-24
Fix TileEntityCrop::onHarvest ItemCrop spawning

CodesCubesAndCrashes:
On the rare occassion that `setCrossCrop(false)` fails, `onHarvest`
should not spawn an ItemCrop anyhow. That would allow the duping of crop
sticks. An example of such a situation is a cross crop that also has a
seed growing on it. Even though that shouldn't happen, the code needs to
watch out for invalid situations like that.

----
#### Screenshot from the #1063 PR, showing the spawning: 
agricraft itemtrowel and cross crops 3


InfinityRaider:
Looks good

Issue #1116, opened 2018-04-04
[1.12.2] Sprinkler pops out if another block is placed next to it


War-Wizard:
## Quick description
What the title says, when a block is placed next to a sprinkler, it pops out as an item.

## Related issues and searches
None that I could find

## Mods and versions used
* Minecraft: 1.12.2
* AgriCraft: 2.12.0-1.12.0-a3
* InfinityLib: 1.12.0

## Steps to replicate the issue
1. Place a sprinkler on a channel
2. Place a block next to it

## Expectations and explanations
The sprinkler should stay where it is as long as it is on a channel.

## Your environment
* How are you playing?
  * Operating System: Windows 10
  * Launcher: Twitch / Vanilla
  * Modpack: Modern Skyblock 3 / Tested stand-alone
* What kind of server:
  * [x] Singleplayer
  * [ ] Multiplayer
* Game Modes:
  * [x] Survival
  * [x] Creative
* Save File:
  * [x] Brand new save
  * [ ] Only an existing save

## Links to crash logs, stack traces, etc
None

## Screenshots (or videos)
Sorry, I'll skip this, should be a video.

PR #1118, opened 2018-04-04
Fix for issue #1116

War-Wizard:
The reason why the sprinkler pops out it's because `canBlockStay` is checking the class of the block above for an interface that's used only in tile entities, so it will never be true. I changed to check for `BlockWaterChannel` which is the class checked for in `canPlaceBlockAt`.

### DIFF
--- a/src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockSprinkler.java
+++ b/src/main/java/com/infinityraider/agricraft/blocks/irrigation/BlockSprinkler.java
@@ -1,6 +1,5 @@
 package com.infinityraider.agricraft.blocks.irrigation;
 
-import com.infinityraider.agricraft.api.v1.misc.IAgriFluidComponent;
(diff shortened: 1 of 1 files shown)


War-Wizard:
No prob, as soon as I get home I'll fix that (in about 5-6 hours).

Quick question: then, shouldn't the player be able to place a sprinkler on any block whose tile entity implements the appropriate interface?

InfinityRaider:
Yes

War-Wizard:
The API compatibility is now fixed, check it whenever you can please.

Regarding the other topic, should I open another issue to talk about it? Right now the sprinkler won't even connect to anything but a BlockWaterChannel, regardless of the API interface implemented, seems like that it's not the intended behaviour.

Maybe `IAgriConnectable.canConnectTo()` should be the check for wether the sprinkler is allowed to be placed? Right now it seems like that method is unused anywhere but in its definition.

PR #1152, opened 2018-06-29
Fix bad import in IAgriSoil

hedgehog1029:
This PR fixes a bad import in IAgriSoil that causes crop sticks to not work correctly on newer Forge builds. This fixes some symptoms of #1115


InfinityRaider:
Thanks!

hedgehog1029:
@EmosewaPixel while you're waiting, you can always add AgriCraft as an override, as I have in my pack [here](https://github.com/general-programming/mc-summer-2018/blob/master/overrides/mods/agricraft-2.12.0-1.12.0-a4.jar). You can use that file, or if you're concerned about its validity you can build from master yourself.

EmosewaPixel:
CurseForge usually doesn't allow that for mods that are already available on CF.

PR #1185, opened 2018-09-07
Update InfinityLib

RlonRyan:
In yet another failed attempt as to fix that annoying already constructing vertices error.


RlonRyan:
Does this need to be closed?

InfinityRaider:
You can update InfinityLib to its final commit on the 1.12 AgriCraft branch.
For now I say we close 1.12 though, as the focus is shifting towards 1.16, and I can't be bothered maintaining multiple versions.

PR #1230, opened 2019-06-23
Add support for EnderIO Farming Stations to harvest AgriCraft crops

rjb3977:
Just as the title says. All this does is add a new "Farmer Joe" to EnderIO's registry which specifically handles AgriCraft's crop block. This doesn't support planting of crops, only picking of pre-planted ones.


dionney:
It works once, then it rips the sticks and lose the seeds.

PR #1234, opened 2019-07-21
Add zh_TW.lang support, and modify some method.

am910021:
Add the "Harvest drops seed" setting to the AgriCraft Config. 
The function of dropping seeds when the harvest is restored.

Add plant name to TileEntityCrop->addDisplayInfo method.
Add plant and soil localization to the addDisplayInfo method.
Add zh_TW.lang support, The main part comes from zh_CN.lang


RlonRyan:
I've manually merged this into my development branch thank you.

PR #1288, opened 2021-01-23
Dev agriseed

InfinityRaider:
Rework the AgriSeed implementation.

All seeds will get converted to the AgriCraft dynamic agri_seed item, the behaviour of which is defined by its NBT.
Since the genome, which is encoded in the nbt, defines the plant species, there can be no inconsistencies between the plant following from the seed item, and the plant following from the genome.

This PR also contains a rework for the rendering of seeds, which now pulls from the models directly, instead of baking a separate model from a texture.

Finally, some bug fixes regarding genome (de)serialization, and the addition of some utility methods for easy conversion between AgriSeed, ItemStack, IAgriGenome, and IAgriPlant.


src/main/java/com/infinityraider/agricraft/api/v1/AgriApi.java
```diff
@@ -190,10 +174,20 @@ public static IAgriStatRegistry getStatRegistry() {
...
         return AgriApi.CONNECTOR.connectFertilizerRegistry();
     }
 
+    @Nonnull
+    public static ItemStack seedToStack(AgriSeed seed, int amount) {
+        return AgriApi.CONNECTOR.seedToStack(seed, amount);
+    }
+
+    @Nonnull
+    public static ItemStack plantToSeedStack(IAgriPlant plant, int amount) {
```
RlonRyan, line 187: Is it meaningful to go from plant to an itemstack without a genome?

src/main/java/com/infinityraider/agricraft/api/v1/AgriApi.java
```diff
@@ -190,10 +174,20 @@ public static IAgriStatRegistry getStatRegistry() {
...
      */
     @Nonnull
-    public static IAgriAdapterizer<IAgriFertilizer> getFertilizerRegistry() {
+    public static IAgriAdapterizer<IAgriFertilizer> getFertilizerAdapterizer() {
         return AgriApi.CONNECTOR.connectFertilizerRegistry();
     }
 
+    @Nonnull
+    public static ItemStack seedToStack(AgriSeed seed, int amount) {
+        return AgriApi.CONNECTOR.seedToStack(seed, amount);
```
RlonRyan, line 183: Should this have been a method on the AgriSeed class?

RlonRyan: I guess if we are divorcing the api from the minecraft code this makes sense.

src/main/java/com/infinityraider/agricraft/api/v1/AgriApi.java
RlonRyan: i.e. should this be genomeToSeedStack?

src/main/java/com/infinityraider/agricraft/api/v1/genetics/IAgriGeneRegistry.java
```diff
@@ -1,12 +1,20 @@
...
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.Optional;
 
 public interface IAgriGeneRegistry extends IAgriRegistry<IAgriGene<?>> {
+    /**
+     * @return The gene responsible for the plant species
+     */
+    @Nonnull
+    IAgriGene<IAgriPlant> getPlantGene();
```
RlonRyan, line 16: Shouldn't this method take an input if it's a registry? As in doesn't it need to do a registry lookup to get the associated value?

src/main/java/com/infinityraider/agricraft/api/v1/genetics/IAgriGenome.java
```diff
@@ -29,6 +30,13 @@
  *    resulting in properties which differ slightly from its parents
  */
 public interface IAgriGenome extends IAgriStatProvider {
+    /**
+     * @return the plant species for this genome
+     */
+    default IAgriPlant getPlant() {
+        return this.getGenePair(AgriApi.getGeneRegistry().getPlantGene()).getTrait();
```
RlonRyan, line 37: Won't this always get the same plant?

src/main/java/com/infinityraider/agricraft/api/v1/seed/AgriSeed.java
```diff
@@ -56,55 +51,26 @@ public ItemStack toStack() {
...
-                .orElseGet(CompoundNBT::new);
-        // Write the genome to the tag.
-        CompoundNBT geneTag = new CompoundNBT();
-        this.genome.writeToNBT(geneTag);
-        tag.put("genome", geneTag);
-        // Return a new stack.
-        ItemStack ret = new ItemStack(stack.getItem(), size);
-        ret.setTag(tag);
-        return ret;
+        return AgriApi.seedToStack(this, size);
```
RlonRyan, line 54: I have a theory this could loose the genome but I could be wrong.

src/main/java/com/infinityraider/agricraft/content/core/BlockCropSticks.java
```diff
@@ -275,8 +275,8 @@ public ActionResultType onBlockActivated(BlockState state, World world, BlockPos
             return ActionResultType.PASS;
         }
         // Fertilization
-        if (AgriApi.getFertilizerRegistry().hasAdapter(heldItem)) {
-            return AgriApi.getFertilizerRegistry().valueOf(heldItem)
+        if (AgriApi.getFertilizerAdapterizer().hasAdapter(heldItem)) {
+            return AgriApi.getFertilizerAdapterizer().valueOf(heldItem)
```
RlonRyan, line 279: The word adapterizer bothers me somehow, even though it's correct. I liked my registries.

src/main/java/com/infinityraider/agricraft/content/core/TileEntityCropSticks.java
```diff
@@ -585,7 +585,7 @@ public boolean hasSeed() {
 
     @Override
     public Optional<AgriSeed> getSeed() {
-        return this.getGenome().flatMap(genome -> Optional.ofNullable(this.hasPlant() ? new AgriSeed(this.getPlant(), genome) : null));
+        return this.getGenome().map(AgriSeed::new);
```
RlonRyan, line 588: So much cleaner.

src/main/java/com/infinityraider/agricraft/handler/ItemToolTipHandler.java
```diff
@@ -29,37 +29,7 @@ public static ItemToolTipHandler getInstance() {
...
-//                final int minAmount = tag.getInteger(IAgriHarvestProduct.PRODUCT_MIN_TAG);
-//                final int maxAmount = tag.getInteger(IAgriHarvestProduct.PRODUCT_MAX_TAG);
-//                final double chance = tag.getDouble(IAgriHarvestProduct.PRODUCT_CHANCE_TAG);
-//                event.getToolTip().add(ChatFormatting.GRAY + "Chance: " + DecimalFormat.getPercentInstance().format(chance));
-//                event.getToolTip().add(ChatFormatting.GRAY + "Min. Amount: " + minAmount);
-//                event.getToolTip().add(ChatFormatting.GRAY + "Max. Amount: " + maxAmount);
-//            }
-//        }
-//    }
+    private ItemToolTipHandler() {}
```
RlonRyan, line 32: Is this file still needed?

RlonRyan (commented):
Looks good, but there is something about the genome registry that seems a little bit off to me.

src/main/java/com/infinityraider/agricraft/api/v1/genetics/IAgriGeneRegistry.java
InfinityRaider: No, this is a quick method to fetch the gene responsible for the plant species. It is always the same, and is not allowed to be removed from the registry.

src/main/java/com/infinityraider/agricraft/api/v1/genetics/IAgriGenome.java
InfinityRaider: No, the genome is not one static instance. The plant gene is, but the value for that gene on the genome (the gene pair) varies.

src/main/java/com/infinityraider/agricraft/api/v1/seed/AgriSeed.java
InfinityRaider: It shouldn't if correctly implemented.

src/main/java/com/infinityraider/agricraft/content/core/BlockCropSticks.java
InfinityRaider: It was confusing me a bit as their data structures are different.

src/main/java/com/infinityraider/agricraft/handler/ItemToolTipHandler.java
InfinityRaider: Useful for debugging

InfinityRaider:
> Looks good, but there is something about the genome registry that seems a little bit off to me.

There is no genome registry, genomes are basically data structures carrying value instances for a set of predefined genes. These genes are registered in the gene registry

src/main/java/com/infinityraider/agricraft/api/v1/AgriApi.java
InfinityRaider: Yes that is the reason for this intermediate method.
If we do scrap the AgriSeed class, this method goes as well.

src/main/java/com/infinityraider/agricraft/api/v1/AgriApi.java
InfinityRaider: You have a valid point

src/main/java/com/infinityraider/agricraft/api/v1/genetics/IAgriGeneRegistry.java
InfinityRaider: To clarify, this is not a registry accessing method, this is an additional utility method to fetch an often used registry element which will always be in the registry without having to do a lookup.

PR #1301, opened 2021-05-03
Update lang file for immersive engineering compatibility

nbrichau:


InfinityRaider:
As always, thanks a lot!

PR #1307, opened 2021-05-14
Add Simple Farming compatibility

nbrichau:


InfinityRaider:
Can you put the textures in the plants subdirectory within textures, and then in another with the mod id? So textures/plant/<mod>/

nbrichau:
I moved the textures files int their directory. I also moved all botania files (models, seed and plant textures) in their proper directory.

I also noticed that their are textures from 1.7/1.12 that are useless (ars magica, thaumcraft), should I remove them too in a future pull-request or add a commit to this one ?

PR #1310, opened 2021-05-18
Added greenhouse world generation

KYPremco:
Implemented the greenhouse world generation from 1.7
Needs infinityLib 1.12.1 to work


InfinityRaider:
This is only the one without irrigation, the one with irrigation is missing?

KYPremco:
Yes this is only the one without irrigation

PR #1313, opened 2021-05-21
Allow custom particles for the plants

nbrichau:


src/main/java/com/infinityraider/agricraft/api/v1/plant/IAgriPlant.java
```diff
@@ -258,6 +259,11 @@ void getClipProducts(@Nonnull Consumer<ItemStack> products, @Nonnull ItemStack c
     @Nonnull
     ResourceLocation getSeedModel();
 
+    /**
+     * @return The particle effect for the plant
+     */
+    List<AgriParticleEffect> getParticleEffects();
```
InfinityRaider, line 265: No referencing to external classes in the API please.
Only Vanilla, Forge, and API classes.

src/main/java/com/infinityraider/agricraft/content/core/BlockCropPlant.java
```diff
@@ -173,4 +179,34 @@ public ActionResultType onBlockActivated(BlockState state, World world, BlockPos
...
+        if (!optional.isPresent()) {
+            return;
+        }
+        IAgriCrop crop = optional.get();
+        if (crop.hasPlant() && crop.isMature()) {
+            List<AgriParticleEffect> particleEffects = crop.getPlant().getParticleEffects();
+            for (AgriParticleEffect particleEffect : particleEffects) {
+                double deltaX = (rand.nextBoolean() ? 1 : -1) * particleEffect.getDeltaX() * rand.nextDouble();
+                double deltaY = (rand.nextBoolean() ? 1 : -1) * particleEffect.getDeltaY() * rand.nextDouble();
+                double deltaZ = (rand.nextBoolean() ? 1 : -1) * particleEffect.getDeltaZ() * rand.nextDouble();
```
InfinityRaider, line 195: Put these in the the final if, as it is not always reached, in which cases these operations are done for nothing.

src/main/java/com/infinityraider/agricraft/impl/v1/plant/NoPlant.java
```diff
@@ -193,6 +194,11 @@ public ResourceLocation getSeedModel() {
         return this.seed_model;
     }
 
+    @Override
+    public List<AgriParticleEffect> getParticleEffects() {
+        return new ArrayList<>();
```
InfinityRaider, line 199: Use Collections.emptyList(), or ImmutableList.of(), no need to make a new object to which people could be adding stuff.

InfinityRaider (changes requested):
This is going to need modifications, instead of using a getter in IAgriPlant, make a method spawnparticles() instead and pass it the IAgriCrop object.
Call this method from the block class, and love the current implementation to JsonPlant. The implementation in NoPlant can be empty

nbrichau:
Hope the changes are good to you.
I used an IAgriStage and not an IAgriCropPlant in IAgriCore#spawnParticles because receiving a stage is enough.

src/main/java/com/infinityraider/agricraft/api/v1/plant/IAgriPlant.java
```diff
@@ -260,9 +261,14 @@ void getClipProducts(@Nonnull Consumer<ItemStack> products, @Nonnull ItemStack c
...
-     * @return The particle effect for the plant
+     * Spawn custom particles if the current growth stage allow it.
+     *
+     * @param world the curent world of the plant.
+     * @param pos the current position of the plant.
+     * @param rand a random for use in rng.
+     * @param stage the current growth stage of the plant.
      */
-    List<AgriParticleEffect> getParticleEffects();
+    void spawnParticles(World world, BlockPos pos, Random rand, IAgriGrowthStage stage);
```
InfinityRaider, line 271: Just pass the IAgriCrop instance (and maybe the Random object), the crop instance gives access to all these parameters and more as well.

InfinityRaider:
> 
> 
> Hope the changes are good to you.
> I used an IAgriStage and not an IAgriCropPlant in IAgriCore#spawnParticles because receiving a stage is enough.

The point of passing in the crop instance is to give people who implement plants via the API all the liberty they might require.
There is no reason to code in limitations.

nbrichau:
Ok, I understand.

InfinityRaider (approved):

PR #1314, opened 2021-05-22
Remove useless files

nbrichau:
I removed Thaumcraft & Ars Magica 2 textures, and the mcmod.info file.

I found also entity textures that I don't feel like they are used. But I didn't remove them as I was not sure. You tell me if I can.


InfinityRaider:
I guess I could fetch them from the 1.12 branch if I'd ever need them again.

nbrichau:
Well, Thaumcraft and Ars Magica 2 didn't update after 1.12 et never will, so you'll never have to fetch them.

darthvader45:
Well, since TeamCoFH has the ability to dev Thaumcraft w/ Azanor having creative control over where it goes, it's possible Thaumcraft could eventually update past 1.12.  It's just not high on the priority list due to other stuff, including life.

nbrichau:
King Lemming also said that we should assume it won't come out.

darthvader45:
Yes, though he has teased things in the past.  Never say never, as they say.

InfinityRaider:
This is not the place to discuss this, i am therefore locking the thread.

PR #1353, opened 2021-06-19
Auto build AgriCraft based on commit v1.12.2

Macleykun:
This pull makes it possible to create a jar based on each commit that's send to the 1.12.2 branch.
By commiting with the message: Release v1.2.3.whatever this CI/CD will create a release in the releases tab with the changelog from the previous version.
Filling in the comment/description field will make the release error out.

This pull has by default set that each Release will be a draft and a pre-release by default. If you wish i can change this behavior.

Each commit will have in this CI/CD a zip with the jars made.

Optional: This CI/CD can also be adjusted to run on pull_requests aswell. So you can quickly test a pull if it works etc.

### DIFF
--- a/.github/workflows/autobuilds.yml
+++ b/.github/workflows/autobuilds.yml
@@ -1,6 +1,6 @@
 name: Autobuild AgriCraft
 
-on: [push, pull-request]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


Macleykun:
Heya, wanted to ask if you had the time to think about this! :D Hope you're doing well!

PR #1354, opened 2021-06-19
Auto build AgriCraft based on commit v1.16

Macleykun:


InfinityRaider:
Thanks, though I need to review this / consider if this fulfills any needs.

Macleykun:
Heya, wanted to ask if you had the time to think about this! :D Hope you're doing well!

InfinityRaider:
Still don't think I need this...

Syndaryl:
I understand you're quite capable of building your own tests, but this would be nice for the folks willing to go with a bleeding edge build to act as a temporary stopgap for crash-on-launch type issues

InfinityRaider:
Yes but I don't want my GitHub repos littered with random releases either.
Building from source isn't that complicated either.

Syndaryl:
As a developer I agree with you re it not being that complicated, but it's quite complicated for people outside our field. It requires installing a lot of software (which is very intimidating, as well as a big download size) and working at the command line, which sadly people don't know much about any more. And half the time the build fails on a new depot because the forge server is burping, and it doesn't explain very well :P 

Releases cluttering up the place is a very painful problem, I agree, can a Travis task delete prior auto-builds so only one "bleeding edge" build is maintained?

InfinityRaider:
That I could consider.

Macleykun:
Do note you never make a Release If you don’t want too.
What initially made is this: https://github.com/Macleykun/AgriCraft/actions/runs/1011058954
Each commit will have a as artefact the build. You won’t see this in your releases tab.
Only if you wish to get something in the releases tab. You make a commit starting with Release v1.2.2 followed by your prefers version numbering. I can remove this last part if you don’t want this.

InfinityRaider:
Besides the optional release, then how is this different than the CI I already have?

Macleykun:
You have any other questions you may have about these pulls?

Macleykun:
Heya, just wanted to send another reminder just in case!

Macleykun:
Deleted the 1.12.2 branch, i assume it's not useful anyways

InfinityRaider:
Since this doesn't have any code changes, no need to keep up to date with master, it won't give problems.
I'll get to this PR when I get to it.

Macleykun:
Heya, i know you've been bussy with life and bug/features. But was wondering if you could take a look at this pull.
It's only really a matter of deciding to have builds in your commits for the people who want to test the build of a given commit. I understand if you don't think this is needed.

PR #1372, opened 2021-07-05
Biomes O' Plenty compatibility

nbrichau:
I also changed your build.gradle because `version` is already appended to `archiveBaseName` when the jar is created.


src/main/java/com/infinityraider/agricraft/plugins/minecraft/JsonPlantCallBackBurn.java
```diff
@@ -0,0 +1,26 @@
...
+    public static JsonPlantCallback getInstance() {
+        return INSTANCE;
+    }
+
+    private JsonPlantCallBackBurn() {
+        super(ID);
+    }
+
+    public void onEntityCollision(@Nonnull IAgriCrop crop, Entity entity) {
+        entity.setFire(1);
```
InfinityRaider, line 24: Perhaps base duration on plant stats?

PR #1389, opened 2021-08-04
Fix rotating flower plant and add language support for MysticalAgriculture and MysticalAgradditions

ArthixFR:
Hi, this PR fixes the rotated flower plant when a MA seeds stage is max and adding MA lang support for displaying correctly the name !


InfinityRaider:
Thanks!

PR #1393, opened 2021-08-15
Update EnderIO API

embeddedt:
Fixes https://github.com/SleepyTrousers/EnderIO/issues/5532

NB: I've never really used this mod, however I set up a quick poppy seed farm in creative and it seems to be farmed correctly by the EnderIO Farming Station.


InfinityRaider:
Thanks.

PR #1449, opened 2022-05-18
New ids and croptopia compatibility

Ketheroth:


src/main/resources/data/agricraft/recipes/agricraft/amethyst_shard.json
```diff
@@ -3,7 +3,7 @@
   "pattern": [
     "ppp",
     "ppp",
-    "pp"
+    "ppp"
```
InfinityRaider, line 6: Good catch!

InfinityRaider:
Looks good, I don't really have any remarks or concerns, good job!

PR #1469, opened 2022-07-13
Update fire.json

Warbringer12:
fixed fire seed inside of the immersive engineering garden cloche to use the correct seed as it was using inferium seed inside of the recipe


InfinityRaider:
Thanks for taking the effort to PR

PR #1484, opened 2022-12-24
Fix multiple mutations with a single output only registering one mutation by changing mutation ID to be based on the file path of the mutation file.

LordPINE:
As the title would suggest, this fixes a bug where only one mutation could be registered for any given product, by replacing the ID used to store the mutation from a product-based one to a file path-based one.

If needed, a change from path to just filename could still be made, though I doubt the necessity.


InfinityRaider:
Thanks!

PR #1537, opened 2024-07-14
Fix Seed Analyzer interactions and minor bugs

oitsjustjose:
AgriCraft currently has two minor bugs with interactions in the Seed Analyzer:

1. When inserting a non-AgriCraft seed into the Seed Analyzer, it will be converted into its AgriCraft equivalent but regardless of how many non-AgriCraft seeds are inserted, the conversion process always sets the stack size to 1, voiding the remainder.

2. Additionally, you are unable to Shift-Click non-AgriCraft seeds into the Seed Analyzer which to many players implies that that item is incompatible with the slot entirely.

This PR introduces a new method which relocates the lambda for converting a non-AgriCraft seed to its AgriCraft equivalent **while retaining the original stack size**. This method will return `ItemStack.EMPTY` if no equivalent is found.

Using this method, Issue **1** from above is resolved as the original stack size is applied. Additionally, by utilizing this method's usage of `ItemStack.EMPTY`, shift-clicking non-AgriCraft seeds has been implemented by verifying that they do in fact have an AgriCraft equivalent thus resolving Issue **2**.


Ketheroth:
Thanks for your fix!
sorry it took so long to get it merged

PR #1550, opened 2024-09-01
[1.20.1] Re-implement mod compat for Botania and Create

unilock:
TODO:
- [ ] Botania mana consumption growth condition / plant modifier
- [ ] Testing!?


common/src/main/java/com/agricraft/agricraft/plugin/botania/BotaniaPlantModifiers.java
```diff
@@ -0,0 +1,62 @@
+package com.agricraft.agricraft.plugin.botania;
+
+public class BotaniaPlantModifiers {
```
InfinityRaider, line 3: What is the point of all of this commented out code?

unilock: Still trying to figure out how to implement it.

From Discord:
> it appears that the "mana consumption" growth condition is presently impossible to implement on 1.20.1 without editing AgriCraft's implementation details due to the available growth requirements being hardcoded in `AgriRequirement.CODEC`
> thus a plugin cannot add its own custom growth requirement by itself

InfinityRaider: That's fine.
In this case I'd prefer not adding the growth rea, adding a TODO comment somewhere relevant and opening an issue for it instead of adding an entire class of commented code.

Ketheroth:
I implemented botania's compatibility in https://github.com/AgriCraft/AgriCraft/commit/3f2dfd04b0922f6dc486e9d1519f6aa4287b3066
I added you as a co-author as I took some of your code and didn't want your work to go to waste

PR #1551, opened 2024-09-02
[1.20.1] Fix planting with Override vanilla farming ON and Plant outside crop sticks OFF

PiotrO15:
Fixes the ability to plant seeds outside crop sticks when Override vanilla farming is set to ON and Plant outside crop sticks is OFF by blocking item use if it's considered an AgriCraft seed and it's used on AgriCraft soil. 

NOTE: this will make it impossible to plant sugar cane on sand (since sugar cane is an AgriCraft seed and sand is an AgriCraft soil). Ideally there could be an additional setting or a json file that lists items except from this change.


InfinityRaider:
Regarding sugarcane, the json used to have a configuration option to handle these exact cases: 
https://github.com/AgriCraft/AgriPlants/blob/master/minecraft/plants/sugarcane_plant.json#L16

Note sure if this survived in the newer versions, if not, readding this would be my preferred solution.

Ketheroth:
I'm gonne close this pr as I've implemented it according to infinityraider's comment in the commit 65e697c77d2463780ddb55816e7facc8a8d3ef6d
I've added you as a co-author.

PR #1562, opened 2024-11-22
[1.21] Only register config screen on client

unilock:
Fixes a startup crash on dedicated servers

Also, I tidied up the buildscript a bit


unilock:
#1565 includes this

PR #1567, opened 2024-12-06
[1.20.1] Industrial Foregoing compat

PiotrO15:
This adds support for Industrial Foregoing plant gatherer which works in a similar way to Botania compat and PNC:R compat (#1564)


Xplodin:
will this ever be released? Could definitely use the compatibility in my modpack

PR #1572, opened 2025-01-25
[1.20.1] Avoid attempting to add compat datapacks / resourcepacks that do not exist

unilock:
Fixes incompatibility with [lazyyyyy](https://modrinth.com/mod/lazyyyyy) (https://github.com/SettingDust/lazyyyyy/issues/29#issuecomment-2613782048)

### DIFF
--- a/fabric/src/main/java/com/agricraft/agricraft/fabric/AgriCraftFabric.java
+++ b/fabric/src/main/java/com/agricraft/agricraft/fabric/AgriCraftFabric.java
@@ -70,7 +70,7 @@ public class AgriCraftFabric implements ModInitializer {
 		FabricLoader.getInstance().getModContainer("agricraft").ifPresent(agricraft -> {
 			for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
 				String modid = mod.getMetadata().getId();
-				if (!modid.equals("agricraft") && !modid.equals("minecraft") && agricraft.findPath("datapacks/"+modid).isPresent() && agricraft.findPath("resourcepacks/"+modid).isPresent()) {
+				if (!modid.equals("agricraft") && !modid.equals("minecraft")) {
 					// deprecated methods are used to avoid using Fabric API internals
 					if (agricraft.findPath("datapacks/"+modid).isPresent()) {
 						ResourceManagerHelper.registerBuiltinResourcePack(new ResourceLocation("builtin", "agricraft_datapacks_" + modid), "datapacks/" + modid, agricraft, CoreConfig.enablePacksByDefault);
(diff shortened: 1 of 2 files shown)


fabric/src/main/java/com/agricraft/agricraft/fabric/AgriCraftFabric.java
```diff
@@ -70,10 +70,10 @@ public void onInitialize() {
 		FabricLoader.getInstance().getModContainer("agricraft").ifPresent(agricraft -> {
 			for (ModContainer mod : FabricLoader.getInstance().getAllMods()) {
 				String modid = mod.getMetadata().getId();
-				if (!modid.equals("minecraft") && !modid.equals("agricraft")) {
+				if (!modid.equals("agricraft") && !modid.equals("minecraft") && agricraft.findPath("datapacks/"+modid).isPresent() && agricraft.findPath("resourcepacks/"+modid).isPresent()) {
```
SettingDust: `agricraft.findPath(“datapacks/”+modid).isPresent() && agricraft.findPath(“resourcepacks/”+modid).isPresent() ` The conditions should be on the following registration statement and not here

SettingDust (changes requested):
it's some log spam instead of incompatibility :)

fabric/src/main/java/com/agricraft/agricraft/fabric/AgriCraftFabric.java
unilock: In place of `CoreConfig.enablePacksByDefault`? That would defeat the point of that config option...

SettingDust: wrap the registerBuiltinResourcePack

unilock: Like so?

```java
if (!modid.equals("agricraft") && !modid.equals("minecraft")) {
	if (agricraft.findPath("datapacks/"+modid).isPresent()) {
		ResourceManagerHelper.registerBuiltinResourcePack(new ResourceLocation("builtin", "agricraft_datapacks_" + modid), "datapacks/" + modid, agricraft, CoreConfig.enablePacksByDefault);
	}
	if (agricraft.findPath("resourcepacks/"+modid).isPresent()) {
		ResourceManagerHelper.registerBuiltinResourcePack(new ResourceLocation("builtin", "agricraft_resourcepacks_" + modid), "resourcepacks/" + modid, agricraft, CoreConfig.enablePacksByDefault);
	}
}
```

The thing is, all (current) AgriCraft datapacks require a corresponding resourcepack, so I figured it would be OK to fail (or pass) hard if one or the other wasn't present.

SettingDust: Ya. Like so. Will work in more situation anyway.

unilock: I guess that does make it match the implementation on Forge...

PR #1573, opened 2025-02-07
Fix invalid pamhc2crops mutations

Minerofmillions:
There were a few typos (primarily pahmhc2crops instead of pamhc2crops) that caused 24 crops to not have valid mutations.


Ketheroth:
closing in favor of #1584

